PR #5, opened 2023-09-18
chore(pkg): Fix module references

simar7:


simar7:
@nikpivkin could you take a look at reviewing this?

nikpivkin:
LGTM
Should we change the [user](https://github.com/aquasecurity/trivy-policies/blob/main/.github/workflows/release.yaml#L7) in the workflow in this PR?

simar7:
> LGTM
> Should we change the [user](https://github.com/aquasecurity/trivy-policies/blob/main/.github/workflows/release.yaml#L7) in the workflow in this PR?

nice catch, I changed to `aqua-bot`.

PR #26, opened 2023-10-24
chore(rules): dedupe AVD-AWS-0180

simar7:
Fixes: https://github.com/aquasecurity/trivy/issues/5065


nikpivkin:
@simar7 We must keep the `no_public_db_access.cf.go` and `no_public_db_access.tf.go` files, since they are used in the Rego rule. https://github.com/aquasecurity/trivy-policies/blob/main/rules/cloud/policies/aws/rds/disable_public_access.rego#L22-L25

Only RDS instances are checked in the Rego rule, but not RDS cluster instances as is done in the Go rule.  https://github.com/aquasecurity/trivy-policies/blob/main/rules/cloud/policies/aws/rds/no_public_db_access.go#L39-L50

simar7:
> @simar7 We must keep the `no_public_db_access.cf.go` and `no_public_db_access.tf.go` files, since they are used in the Rego rule. https://github.com/aquasecurity/trivy-policies/blob/main/rules/cloud/policies/aws/rds/disable_public_access.rego#L22-L25
> 
> Only RDS instances are checked in the Rego rule, but not RDS cluster instances as is done in the Go rule. https://github.com/aquasecurity/trivy-policies/blob/main/rules/cloud/policies/aws/rds/no_public_db_access.go#L39-L50

good catch, I merged your PR.

PR #28, opened 2023-10-24
chore: remove docs

nikpivkin:


simar7:
before we can merge this is in, we need to update the trivy-iac repo with the new docs and also update the avd-generator to use the new docs from trivy-iac repo https://github.com/aquasecurity/avd-generator/blob/f0d351962e2e5b61fbdd32b0426f4da4a6acecdd/docGen/main.go#L49

simar7:
Closing this as `trivy-iac` will get merged into Trivy so we'll need to keep the docs (and their generation) here. More details here https://github.com/aquasecurity/trivy-policies/pull/39

PR #39, opened 2023-11-21
refactor(deps): Replace `trivy-iac/pkg` with `defsec/pkg`

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/5626


simar7:
@nikpivkin could you take a look at reviewing this?

PR #46, opened 2023-12-13
feat(aws): improve AVD-AWS-0057 rule

nikpivkin:
ARNs that contain a wildcard in the object key are now allowed, as such an ARN has fewer object access privileges than an ARN that identifies all objects in the bucket. 

See https://github.com/aquasecurity/trivy/issues/5114

### DIFF
--- a/checks/cloud/aws/iam/no_policy_wildcards.go
+++ b/checks/cloud/aws/iam/no_policy_wildcards.go
@@ -6,18 +6,12 @@ import (
 	"strings"
 
 	"github.com/aquasecurity/defsec/pkg/framework"
-
-	"github.com/aquasecurity/defsec/pkg/providers/aws/iam"
-	"github.com/aquasecurity/defsec/pkg/severity"
-
-	"github.com/aquasecurity/defsec/pkg/state"
-
-	"github.com/aquasecurity/defsec/pkg/scan"
-
-	"github.com/aquasecurity/trivy-policies/pkg/rules"
-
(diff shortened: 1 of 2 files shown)


simar7 (approved):

checks/cloud/aws/iam/no_policy_wildcards.go
```diff
@@ -165,3 +159,21 @@ func checkStatement(src iam.Document, statement iamgo.Statement, results scan.Re
...
+	if len(arnParts) != 6 || arnParts[2] != "s3" {
+		return false
+	}
+
+	resourceParts := strings.SplitN(arnParts[5], "/", 2)
+	if len(resourceParts) != 2 {
+		return false
+	}
+
+	return !strings.ContainsRune(resourceParts[0], '*') && strings.ContainsRune(resourceParts[1], '*')
```
simar7, line 177: this logic seems a little brittle to me, don't you think?

nikpivkin: I don't think the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html) format will change. Does it make sense to use the [aws-sdk-go](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/arn/arn.go) package for such a simple check?

simar7: Fair enough

PR #53, opened 2024-01-10
fix: fail for irrelevant resources

itaysk:
reported in https://github.com/aquasecurity/trivy/issues/4922 since the failRootGroupId rule is given a default value, it doesn't fail the deny rule and results in a false detection. added an explicit check for the failRootGropuId truthiness to resolve.

also fixed the sad path test
however I'm not sure how we're supposed to run the tests and if it's indeed broken

### DIFF
--- a/checks/kubernetes/general/runs_with_a_root_primary_or_supplementary_GID.rego
+++ b/checks/kubernetes/general/runs_with_a_root_primary_or_supplementary_GID.rego
@@ -59,6 +59,7 @@ failRootGroupId {
 }
 
 deny[res] {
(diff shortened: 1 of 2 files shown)


itaysk:
p.s I would expect the selector subtype to have prevent this bug, no?

simar7:
> p.s I would expect the selector subtype to have prevent this bug, no?

k8s subtypes support was [just added](https://github.com/aquasecurity/defsec/pull/1502), we need to bump the dependency for it to pick it up. I'm taking care of it.

simar7 (approved):

PR #55, opened 2024-01-16
add check if there is actually a package manager in the run command

Morl99:
Simplify rule to use regex instead of splitting

closes https://github.com/aquasecurity/defsec/issues/1256

### DIFF
--- a/checks/docker/update_instruction_alone.rego
+++ b/checks/docker/update_instruction_alone.rego
@@ -24,6 +24,7 @@ deny[res] {
 
 	command = concat(" ", run.Value)
 
+	is_package_manager(command)
 	is_valid_update(command)
 	not update_followed_by_install(command)
 
@@ -31,16 +32,16 @@ deny[res] {
 	res := result.new(msg, run)
 }
 
+package_manager_regex := `(apk)|(apt-get)|(yum)`
+
+is_package_manager(command) {
+	regex.match(package_manager_regex, command)
+}
+
+update_regex := `( update)|( check-update)`
+
 is_valid_update(command) {
-	chained_parts := regex.split(`\s*&&\s*`, command)
-
-	array_split := split(chained_parts[_], " ")
-
-	len = count(array_split)
-
-	update := {"update", "--update"}
-
-	array_split[len - 1] == update[_]
+	regex.match(update_regex, command)
 }
 
 update_followed_by_install(command) {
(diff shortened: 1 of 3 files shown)


Morl99:
The CLA checker does not seem to be aware, that I am a member of the https://github.com/dbsystel organization, and that our organization has signed a CLA.

simar7:
Thanks could you format the rego check with `opa fmt -w`?

Morl99:
> Thanks could you format the rego check with `opa fmt -w`?

Done

I have added a line about that to the `CONTRIBUTING.md`.

checks/docker/update_instruction_alone.rego
```diff
@@ -24,23 +24,24 @@ deny[res] {
...
+is_package_manager(command) {
+	regex.match(package_manager_regex, command)
+}
 
-	update := {"update", "--update"}
+update_regex := `( update)|( check-update)`
 
-	array_split[len - 1] == update[_]
+is_valid_update(command) {
+	regex.match(update_regex, command)
```
nikpivkin, line 44: When using regex, we will not be able to establish that the update command refers to the package manager. Example of a command that causes a false positive:`apt-get build-dep && /bin/sh /scripts/someScript.sh update`. Maybe we should split the command by && and check each part?

nikpivkin:
@simar7 I think we should also check the package manager along with the install command. 
`apt-get update && /bin/sh /scripts/someScript.sh install` doesn't trigger a warning, but this `apt-get update && /bin/sh /scripts/someScript.sh other-cmd` does.  Does it relate to this PR?

Morl99:
> @simar7 I think we should also check the package manager along with the install command. 
> 
> `apt-get update && /bin/sh /scripts/someScript.sh install` doesn't trigger a warning, but this `apt-get update && /bin/sh /scripts/someScript.sh other-cmd` does.  Does it relate to this PR?

Not sure I understand what you mean. Does `apt-get update && /bin/sh /scripts/someScript.sh other-cmd` trigger the rule that is modified in this PR? If yes, that surprises me. If it is a different rule, you should open a separate issue and we should go ahead and merge this one to close the original issue.

simar7:
> > @simar7 I think we should also check the package manager along with the install command.
> > `apt-get update && /bin/sh /scripts/someScript.sh install` doesn't trigger a warning, but this `apt-get update && /bin/sh /scripts/someScript.sh other-cmd` does.  Does it relate to this PR?
> 
> True, that is the other side of the original issue, that we could fix here as well. I will rework the code to also cover this false negative.

Sounds good to me.

PR #58, opened 2024-01-17
fix: remove test-rego from Makefile

candrews:
e339a0d removed the `test-rego` rule so references to it should also be removed.


simar7 (approved):
thanks, we do plan to add the target back soon but it's good to clean it up now, until we do so.

PR #60, opened 2024-01-17
fix(checks): handle `file:` and `multi:` in AVD-DS-005

candrews:
The reverse engineered `Dockerfile` of an image doesn't exactly match the original `Dockerfile`. For example, it doesn't have the original source files names. Instead, it uses `file:<hash> in`: `ADD file:8b8864b3e02a33a579dc216fd51b28a6047bc8eeaa03045b258980fe0cf7fcb3 in /__cacert_entrypoint.sh`

Such commands should not trigger AVD-DS-005.

### DIFF
--- a/checks/docker/add_instead_of_copy.rego
+++ b/checks/docker/add_instead_of_copy.rego
@@ -24,10 +24,20 @@ get_add[output] {
 	args := concat(" ", add.Value)
 
 	not contains(args, ".tar")
+
+	not is_command_with_hash(add.Value, "file:")
+	not is_command_with_hash(add.Value, "multi:")
+
 	output := {
 		"args": args,
 		"cmd": add,
 	}
+}
+
+is_command_with_hash(cmd, prefix) {
+	count(cmd) == 3
+	startswith(cmd[0], prefix)
+	cmd[1] == "in"
 }
 
 deny[res] {
--- a/checks/docker/add_instead_of_copy_test.rego
+++ b/checks/docker/add_instead_of_copy_test.rego
@@ -44,6 +44,24 @@ test_copy_allowed {
 	count(r) == 0
 }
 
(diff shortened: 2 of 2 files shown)


candrews:
@simar7 can you please help me progress this MR?

I'm very eager to eliminate this false positive.

simar7:
@nikpivkin could you take a look?

checks/docker/add_instead_of_copy.rego
```diff
@@ -24,6 +24,10 @@ get_add[output] {
 	args := concat(" ", add.Value)
 
 	not contains(args, ".tar")
+
+	cnt := count(copy.Value)
+	not (cnt == 3 && startswith(add.Value[0], "file:") || startswith(add.Value[0], "multi:") && add.Value[1] == "in")
+
```
nikpivkin, line 30: I don't have permissions to make changes to this branch.

Rego does not have the `&&` and `||` operators. All expressions in a rule are evaluated sequentially, which is equivalent to conjunction.

```suggestion
	not is_command_with_hash(add.Value, "file:")
	not is_command_with_hash(add.Value, "multi:")
```

```rego
is_command_with_hash(cmd, prefix) {
	count(cmd) == 3
	startswith(cmd[0], prefix)
	cmd[1] == "in"
}

```

You can also merge main into your branch to run tests.

nikpivkin: @candrews could you please update [this](https://github.com/aquasecurity/trivy-policies/pull/56) PR as well?

candrews: Thank you for providing the ability to run tests, that's very helpful!

I've updated this MR with your suggestion.

nikpivkin (approved):

simar7 (approved):

PR #65, opened 2024-01-29
fix: add system authenticate bind protection policy

chen-keinan:
Following to security risk raised on [system:authenticate ](https://orca.security/resources/blog/sys-all-google-kubernetes-engine-risk/) adding a new policy to handle this case

### DIFF
--- a/avd_docs/kubernetes/gke/AVD-KSV-01011/docs.md
+++ b/avd_docs/kubernetes/gke/AVD-KSV-01011/docs.md
@@ -1,0 +1,10 @@
+
+Binding to system:authenticate group to any clusterrole or role is a security risk.
+
+### Impact
+<!-- Add Impact here -->
+
+<!-- DO NOT CHANGE -->
+{{ remediationActions }}
+
+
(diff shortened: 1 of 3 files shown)


simar7 (approved):

knqyf263:
I don't recall we have such logic, but can we apply this policy to only GKE clusters?

chen-keinan:
@knqyf263 no we can't , there is an  issue #6031 to pass distribution type as rego input (will be done later). this policy is important for all distribution it is more exposable for GKE

knqyf263:
>this policy is important for all distribution it is more exposable for GKE

IIUC, as it is infinitely unlikely that the `system:authenticated` group will be assigned to an attacker in non-GKE clusters, I'm concerned that the detection of this check as CRITICAL will be noise. This could frighten many EKS and AKS users.

chen-keinan:
> > this policy is important for all distribution it is more exposable for GKE
> 
> IIUC, as it is infinitely unlikely that the `system:authenticated` group will be assigned to an attacker in non-GKE clusters, I'm concerned that the detection of this check as CRITICAL will be noise. This could frighten many EKS and AKS users.

the chances for `system:authenticated` group to by bind to a Role as the same for all (GKE and non GKE) as this can be done by mistake, the problem with GKE is that the generated which can be easily abused

knqyf263:
@chen-keinan Any comments?

chen-keinan:
> @chen-keinan Any comments?

I'll change the severity to Low

knqyf263:
But it is critical for GKE.

- GKE: CRITICAL
- Non-GKE: LOW

We want to achieve it somehow.

PR #72, opened 2024-02-07
feat(checks): Add check to detect Leaky Vessels exploit

simar7:
This PR adds a check to detect the [leaky vessels exploit](https://github.com/snyk/leaky-vessels-static-detector/) with Trivy.

Also fixes: https://github.com/aquasecurity/trivy-policies/issues/67

Signed-off-by: Simar <simar@linux.com>

### DIFF
--- a/checks/docker/prevent_sys_workdir_mounts.rego
+++ b/checks/docker/prevent_sys_workdir_mounts.rego
@@ -1,0 +1,40 @@
+# METADATA
+# title: "WORKDIR should not be mounted on system dirs"
+# description: "WORKDIR should not be mounted on system directories to avoid container breakouts"
+# scope: package
+# schemas:
+# - input: schema["dockerfile"]
+# related_resources:
+# - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#workdir
+# custom:
+#   id: DS030
+#   avd_id: AVD-DS-0030
+#   severity: HIGH
+#   short_code: avoid-sys-workdir-mounts
+#   recommended_action: "Avoid using system directories to mount WORKDIR"
+#   input:
+#     selector:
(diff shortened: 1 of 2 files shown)


simar7:
@nikpivkin could you help to review this?

checks/docker/prevent_sys_workdir_mounts.rego
```diff
@@ -0,0 +1,40 @@
...
+#   short_code: avoid-sys-workdir-mounts
+#   recommended_action: "Avoid using system directories to mount WORKDIR"
+#   input:
+#     selector:
+#     - type: dockerfile
+package builtin.dockerfile.DS030
+
+import data.lib.docker
+
+sysdirs := {"/proc", "/boot", "/dev", "/initrd", "/lost+found"}
```
itaysk, line 22: best to add a tailing slash to prevent partial matching. for example someone might set workdir to `/development` which is legit but will be flagged by this check as it `startswith` `/dev`

simar7: oh yeah good point, I've updated that and added a test.

simar7:
> @simar7 Does this rule detect [CVE-2024-21626](https://github.com/advisories/GHSA-xr7r-f8xq-vfvv)? If so, the attack is only possible using the `/proc/self/fd/*` path in the `WORKDIR` instruction. Leaky Vessels includes [four](https://snyk.io/blog/leaky-vessels-docker-runc-container-breakout-vulnerabilities/) vulnerabilities.

We want to in general avoid people using system dirs to mount their workdirs. This also prevents the CVE you mentioned but isn't limited to just that.

itaysk (approved):

PR #74, opened 2024-02-08
fix: change system authenticate bind severity to low

chen-keinan:
## Description
change system authenticate bind severity to low
- Related #65


simar7 (approved):
nice!

PR #87, opened 2024-03-02
feat: add DeploymentConfig support

szubersk:
Add support for OpenShift's [DeploymentConfig](https://docs.openshift.com/container-platform/4.12/rest_api/workloads_apis/deploymentconfig-apps-openshift-io-v1.html). DeploymentConfig is an enhanced Deployment.

> Deployment Configs define the template for a pod and manages deploying new images or configuration changes. A single deployment configuration is usually analogous to a single micro-service. Can support many different deployment patterns, including full restart, customizable rolling updates, and fully custom behaviors, as well as pre- and post- deployment hooks. Each individual deployment is represented as a replication controller.
> 
> A deployment is "triggered" when its configuration is changed or a tag in an Image Stream is changed. Triggers can be disabled to allow manual control over a deployment. The "strategy" determines how the deployment is carried out and may be changed at any time. The latestVersion field is updated when a new deployment is triggered by any means.

### DIFF
--- a/checks/kubernetes/advanced/default_namespace_should_not_be_used.rego
+++ b/checks/kubernetes/advanced/default_namespace_should_not_be_used.rego
@@ -20,6 +20,7 @@
 #         - kind: replicaset
 #         - kind: replicationcontroller
 #         - kind: deployment
+#         - kind: deploymentconfig
 #         - kind: statefulset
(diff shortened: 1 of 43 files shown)


chen-keinan (approved):
lgtm 🚀

PR #90, opened 2024-03-05
fix(azure): update remediation for AVD-AZU-0027

nikpivkin:
Fixes https://github.com/aquasecurity/trivy/issues/5739


simar7 (dismissed):

simar7:
Actually one thing we can improve is to add a check to see if docs are up to date when checks are modified in anyway. For instance in this PR we have to regenerate the docs to update them with the new example. 

I've pushed it for now.

simar7 (approved):

PR #97, opened 2024-03-13
fix(checks): check the package manager in AVD-DS-0017

nikpivkin:


nikpivkin:
@simar7 Are there any package managers that need to be added?

simar7 (approved):
lgtm for now

PR #100, opened 2024-03-19
Fix check for public addresses

kidbomb:
According to the check documentation:

> Opening up ports to the public internet is generally to be avoided. You should restrict access to IP addresses or ranges that explicitly require it where possible

However, the check fails even if you try to set the cidr_blocks parameter to a more restrictive range, such as:

```
 resource "aws_security_group_rule" "another_good_example" {
  type = "ingress"
  cidr_blocks = ["1.2.3.4/24"]
 }
```

Running `trivy config .` with the configuration above yields:

```
CRITICAL: Security group rule allows ingress from public internet.
════════════════════════════════════════════════════════════════════════════════════════════════════
Opening up ports to the public internet is generally to be avoided. You should restrict access to IP addresses or ranges that explicitly require it where possible.

See https://avd.aquasec.com/misconfig/avd-aws-0107
────────────────────────────────────────────────────────────────────────────────────────────────────
 main.tf:3
   via main.tf:1-4 (aws_security_group_rule.another_good_example)
────────────────────────────────────────────────────────────────────────────────────────────────────
   1    resource "aws_security_group_rule" "another_good_example" {
   2     type = "ingress"
   3 [   cidr_blocks = ["1.2.3.4/24"]
   4    }

```

### DIFF
--- a/avd_docs/aws/ec2/AVD-AWS-0107/Terraform.md
+++ b/avd_docs/aws/ec2/AVD-AWS-0107/Terraform.md
@@ -9,6 +9,13 @@ Set a more restrictive cidr range
  
 ```
 ```hcl
+ resource "aws_security_group_rule" "another_good_example" {
+ 	type = "ingress"
+ 	cidr_blocks = ["1.2.3.4/24"]
+ }
+ 
+```
+```hcl
 resource "aws_security_group_rule" "allow_partner_rsync" {
   type              = "ingress"
   security_group_id = aws_security_group.….id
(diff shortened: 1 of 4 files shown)


nikpivkin:
@simar7 We check the CIDR for publicity in many checks. Should we fix them?

simar7:
> @simar7 We check the CIDR for publicity in many checks. Should we fix them?

Can you give an example and the proposed improvement?

nikpivkin:
@simar7  I can fix the rest of the checks in a separate PR.

nikpivkin:
@kidbomb Thanks for the PR! We are [bringing](https://github.com/aquasecurity/trivy-checks/pull/257) AVD-AWS-0107 and AVD-AWS-0105 checks in line with AWS CIS Benchmarks, so our changes will fix this.

PR #113, opened 2024-04-18
fix(rego): improve commands parsing

nikpivkin:
Fixes https://github.com/aquasecurity/trivy/issues/6516

Fixed DS017 check, now commands are parsed correctly.

Added a new function `sh.parse_commands` that takes a string representation of a sequence of commands and returns a list of lists of command parts.
For example:
```sh
./opa eval 'sh.parse_commands("apt update; apt install -y nginx")' -f pretty
[
  [
    "apt",
    "update"
  ],
  [
    "apt",
    "install",
    "-y",
    "nginx"
  ]
]
```

We have a lot of checks where we parse commands using regular expressions, which can be unreliable.. Later we can move on to using this function.


nikpivkin:
@simar7 Will be ready to merge after https://github.com/aquasecurity/trivy/pull/6502 for backwards compatibility in users who don't update Trivy.

simar7 (approved):

PR #120, opened 2024-04-26
chore(deps): Fix testcontainers import

simar7:
Use the same ones that Trivy does as the latest has a bug upstream.


go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/aquasecurity/trivy-checks
 
-go 1.22.0
+go 1.22
+
+toolchain go1.22.0
```
simar7, line 5: Need to keep in sync with Trivy due to https://github.com/aquasecurity/trivy/pull/6476#discussion_r1582593900

PR #129, opened 2024-05-14
chore(checks): Add CSPM ID for AVD-AWS-0089

simar7:
Related https://github.com/aquasecurity/trivy/issues/6686


RobinFrcd (approved):

RobinFrcd:
https://avd.aquasec.com/misconfig/aws/s3/avd-aws-0091/: `aws-s3-ignore-public-acls`

simar7:
> https://avd.aquasec.com/misconfig/aws/s3/avd-aws-0091/: `aws-s3-ignore-public-acls`

We already ignore by LongID() which is defined [here](https://github.com/aquasecurity/trivy/blob/9cd5294a21b42ad6aa4149d5c0374fcf80f5cc53/pkg/iac/scan/rule.go#L76)

If this LongID isn't being ignored, we have a different bug to fix that is not related to aliasing.

PR #165, opened 2024-06-26
test: init functional tests

nikpivkin:
This PR adds functional tests for already migrated checks from Go to Rego to prevent regression. The remaining tests will be added as the checks are moved.

### DIFF
--- a/test/rego_checks_test.go
+++ b/test/rego_checks_test.go
@@ -1,0 +1,744 @@
+package test
+
+import (
+	"context"
+	"testing"
+
+	checks "github.com/aquasecurity/trivy-checks"
+	"github.com/aquasecurity/trivy/pkg/iac/providers/aws"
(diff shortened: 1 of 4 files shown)


test/rego_checks_test.go
```diff
@@ -0,0 +1,744 @@
...
+
+	tests := map[string][]testCase{
+		"AVD-AWS-0175": {
+			// TODO: Trivy does not export empty structures into Rego
+			// {
+
+			// 	name:     "No analyzers enabled",
+			// 	input:    state.State{AWS: aws.AWS{AccessAnalyzer: accessanalyzer.AccessAnalyzer{}}},
+			// 	expected: true,
+			// },
```
simar7, line 63: Pardon my memory but did we ever decide how to proceed on this?

nikpivkin: We've discussed this in DM, but we haven't come to a conclusion yet.

simar7 (approved):

PR #173, opened 2024-07-06
test(bundle): Verify bundle usage

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/7031

### DIFF
--- a/scripts/verify-bundle.go
+++ b/scripts/verify-bundle.go
@@ -15,6 +15,7 @@ import (
 
 var bundlePath = "bundle.tar.gz"
 var OrasPush = []string{"--config", "/dev/null:application/vnd.cncf.openpolicyagent.config.v1+json", fmt.Sprintf("%s:application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip", bundlePath)}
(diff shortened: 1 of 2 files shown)


scripts/verify-bundle.go
```diff
@@ -14,6 +15,7 @@ import (
 
 var bundlePath = "bundle.tar.gz"
 var OrasPush = []string{"--config", "/dev/null:application/vnd.cncf.openpolicyagent.config.v1+json", fmt.Sprintf("%s:application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip", bundlePath)}
+var supportedTrivyVersions = []string{"latest", "canary"} // TODO: add more versions
```
nikpivkin, line 18: More versions will be added later?

nikpivkin (approved):

scripts/verify-bundle.go
simar7: I wasn't able to find a way without using Docker's API (requires a token) to be able to fetch tags. We could scrape them but I thought we can start with these first and then improve upon. 

Do you know of any way to get the tags without auth or scrape?

nikpivkin: We can retrieve the tags from GitHub.

PR #210, opened 2024-08-10
Ignore casing in CIDR wildcards

InverseIntegral:
# Problem

Trivy doesn't detect any problem in the following Terraform file:
```terraform
resource "azurerm_network_security_group" "example" {
  security_rule {
    direction                  = "Inbound"
    access                     = "Allow"
    source_address_prefix      = "Internet"
  }
}
```

Changing it to this, however, results in a critical finding:

```terraform
resource "azurerm_network_security_group" "example" {
  security_rule {
    direction                  = "Inbound"
    access                     = "Allow"
    source_address_prefix      = "internet"
  }
}
```

Note the different casing of the `source_address_prefix`.

# Solution

I've adjusted the CIDR checks to include the wildcards regardless of their casing. `internet` and `Internet` should now be treated the same way.

### DIFF
--- a/checks/cloud/azure/network/no_public_ingress_test.go
+++ b/checks/cloud/azure/network/no_public_ingress_test.go
@@ -20,7 +20,7 @@ func TestCheckNoPublicIngress(t *testing.T) {
 		expected bool
 	}{
 		{
-			name: "Security group inbound rule with wildcard source address",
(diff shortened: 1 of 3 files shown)


nikpivkin (approved):
LGTM

simar7 (approved):

PR #218, opened 2024-08-20
chore: init separate package for bundle scripts

nikpivkin:
The docker and test-containers packages are only used for bundle verification. Сreating a separate package for scripts reduce the dependency resolution chore when updating Trivy and reduce dependabot noise.


scripts/go.mod
```diff
@@ -0,0 +1,60 @@
+module github.com/aquasecurity/trivy-checks/bundle
```
simar7, line 1: why does this file (and also the go.sum) have git addition marks? (The plus sign appended to each of the lines).

nikpivkin: it's a new file

simar7: doh 🤦🏼 you're right!

simar7 (approved):

PR #226, opened 2024-08-21
refactor(checks): migrate AWS ec2 to Rego

nikpivkin:
Related PRs:
- #216


nikpivkin:
The error occurred on the Trivy side. Opened a PR with a fix. https://github.com/aquasecurity/trivy/pull/7377

nikpivkin:
@simar7 Trivy now contains Go checks that are not marked as deprecated, so they are not ignored before applying all checks as a result they are duplicated. But this will be fixed if we update trivy-checks in Trivy.

PR #234, opened 2024-08-22
test: initialise tests in each test file

nikpivkin:
Will avoid merge conflicts


simar7 (approved):
it's fine but it's much easier if we have them back in the map as if we ever need to run a subset of tests, for  instance for debugging, it's easier to comment them out.

nikpivkin:
I'll bring it back later.

PR #235, opened 2024-08-22
refactor(checks): migrate AWS IAM to Rego 

nikpivkin:


nikpivkin:
The error occurred on the Trivy side. Opened a PR with a fix https://github.com/aquasecurity/trivy/pull/7376

simar7 (approved):

PR #245, opened 2024-08-27
chore(checks): deprecate some checks

nikpivkin:
1. [AVD-AWS-0134](https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/aws/ssm/avoid_leaks_via_http.go#L13). This check uses a mechanism of custom checks, which are passed terraform blocks directly rather than general state. Because of this, this check cannot be ported to Rego.
2. [AVD-AWS-0175](https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/aws/accessanalyzer/enable_access_analyzer.go) and [AVD-AWS-0169](https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/aws/iam/require_support_role.go). These checks are triggered when there are no resources. The problem is that Trivy does not export empty structures for Rego. Because of this, filtering with selectors does not work and the check is not applied. If you remove the selectors, the check will always work, even when not scanning AWS. These checks are better evaluated in other ways as these resources can be configured in other ways.

### DIFF
--- a/checks/cloud/aws/iam/no_policy_wildcards.go
+++ b/checks/cloud/aws/iam/no_policy_wildcards.go
@@ -49,7 +49,8 @@ var CheckNoPolicyWildcards = rules.Register(
 			Links:               cloudFormationNoPolicyWildcardsLinks,
 			RemediationMarkdown: cloudFormationNoPolicyWildcardsRemediationMarkdown,
 		},
-		Severity: severity.High,
+		Severity:   severity.High,
+		Deprecated: true,
 	},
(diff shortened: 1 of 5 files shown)


knqyf263:
@nikpivkin AVD-AWS-0057 was marked deprecated in this PR, but the reason is not mentioned in the description. Where can I see the discussion?

nikpivkin:
@knqyf263 We discussed this at length with Simar in DM and I think I forgot to add it to the description

knqyf263:
I got it. Can you please update the description?

PR #250, opened 2024-08-30
chore: mark Rego libs as libs

nikpivkin:
Fixes: https://github.com/aquasecurity/trivy/issues/7426

### DIFF
--- a/lib/cloud/aws_trails.rego
+++ b/lib/cloud/aws_trails.rego
@@ -1,3 +1,9 @@
+# METADATA
+# custom:
+#   library: true
+#   input:
+#     selector:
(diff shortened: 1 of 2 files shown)


simar7 (approved):

simar7:
@nikpivkin I just tested this again with trivy main and it seems to have no effect

```
rm -rf ~/Library/Caches/trivy/  && ./trivy --skip-check-update --debug config  <input>

2024/08/31 01:22:47 WARN Check ID is empty file_path=lib/docker/docker.rego
2024/08/31 01:22:47 WARN Check ID is empty file_path=lib/cloud/aws_trails.rego
```

PR #264, opened 2024-10-03
fix(rego): Use correct metadata for libs

simar7:


nikpivkin (approved):

lib/test/lib_test.rego
```diff
@@ -1,6 +1,3 @@
-# METADATA
-# custom:
-#   library: true
 package lib.test
```
nikpivkin, line 1: Would that get rid of the message that there are no selectors?

simar7: Actually it might be just enough to rename the file as _test.rego based on my testing. I'll give it another try to confirm.

simar7: It seems to be the case, I've reverted this back.

simar7:
Closing in favor of https://github.com/aquasecurity/trivy/pull/7638

PR #268, opened 2024-10-09
fix(azure): properly check ports in AVD-AZU-0058 and AVD-AZU-0050

nikpivkin:
Fixes: https://github.com/aquasecurity/trivy/issues/7681

### DIFF
--- a/checks/cloud/azure/network/disable_rdp_from_internet.rego
+++ b/checks/cloud/azure/network/disable_rdp_from_internet.rego
@@ -49,4 +49,7 @@ deny contains res if {
 	)
 }
 
-port_range_includes(from, to, port) if from <= port <= to
+port_range_includes(from, to, port) if {
+	from.value <= port
+	port <= to.value
+}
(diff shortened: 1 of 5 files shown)


checks/cloud/azure/network/disable_rdp_from_internet.rego
```diff
@@ -49,4 +49,7 @@ deny contains res if {
 	)
 }
 
-port_range_includes(from, to, port) if from <= port <= to
+port_range_includes(from, to, port) if {
+	from.value <= port
+	port <= to.value
+}
```
simar7, line 55: I'm curious is this valid:

```if from.value <= port <= to.value```?

simar7 (approved):

checks/cloud/azure/network/disable_rdp_from_internet.rego
nikpivkin: This kind of expression doesn't work in Rego: https://play.openpolicyagent.org/p/2RS5HYLT5d

PR #271, opened 2024-10-14
chore: store examples in YAML

nikpivkin:
In this PR: 
- Terraform and CloudFormation examples moved from `cf.go` and `tf.go` to `yaml`, 
- all checks now refer to the yaml examples
- added command for formatting examples

Although all the examples for a single check are in a single file, we need to reference it from multiple check fields. I suggest adding a new `examples` field that will point to a single file, while maintaining support for the current method of referencing examples for backwards compatibility.

Old example files will be deleted in another PR as they are still referenced by Go checks.


nikpivkin:
Will be ready for review after the release of Trivy v0.57

simar7:
One thing to keep in mind with the examples is the avd generator. I believe it uses the go files as well. Could you confirm that @nikpivkin?

nikpivkin:
@simar7 I'll check, but as far as I remember avd-generator uses generated md files.

nikpivkin:
avd-generator uses only generated `.md` files https://github.com/aquasecurity/avd-generator/blob/main/docGen/defsec.go#L151-L161

nikpivkin:
Maybe we should change the file schema to the following?
```yaml
id: avdid
examples:
   terraform:
   cloudformation:
```

This will allow us to associate examples and checks without using the `trivy` and `trivy-checks` packages to run the tests. To do this now, we need to load all the checks and then load all the examples based on the check metadata.

simar7:
> Maybe we should change the file schema to the following?
> 
> ```yaml
> id: avdid
> examples:
>    terraform:
>    cloudformation:
> ```
> 
> This will allow us to associate examples and checks without using the `trivy` and `trivy-checks` packages to run the tests. To do this now, we need to load all the checks and then load all the examples based on the check metadata.

It makes sense to me but we should make an announcement regarding this change as we will be changing the metadata schema. I suggest let's do it in a separate PR.

simar7 (approved):
🙏🏼

PR #277, opened 2024-10-16
feat(checks): Deprecate remaining PSP checks

simar7:
Close: https://github.com/aquasecurity/trivy/issues/5541

### DIFF
--- a/checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
+++ b/checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
@@ -9,6 +9,9 @@
 # custom:
 #   id: KCV0013
 #   avd_id: AVD-KCV-0013
+#   frameworks:
+#     k8s-cis-1.23:
+#       - "1.23"
 #   severity: LOW
 #   short_code: ensure-admission-control-plugin-security-context-deny-is-set-if-pod-security-policy-is-not-used
 #   recommended_action: "Edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml on the Control Plane node and set the --enable-admission-plugins parameter to include SecurityContextDeny, unless PodSecurityPolicy is already in place."
--- a/checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
+++ b/checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
@@ -11,7 +11,9 @@
 #   avd_id: AVD-KCV-0013
 #   frameworks:
 #     k8s-cis-1.23:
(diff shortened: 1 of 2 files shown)


simar7:
@nikpivkin since you recently checked the CIS benchmarks, should they be updated as well as some of them check for this check as can be seen here: https://github.com/aquasecurity/trivy-checks/blob/6f4dbd8399cecf9d65a76232a6adb5115a0e3ba0/pkg/specs/compliance/k8s-cis-1.23.yaml#L285-L293

nikpivkin:
@simar7 You mean update the checks to align with the benchmarks?

simar7:
> @simar7 You mean update the k8s checks to align with the benchmarks?

I think we will need to keep that check as we only target 1.23 which still has it. Unless we update to target a newer version of k8s.

nikpivkin:
@simar7 In such a situation, we specify a frameworks field for cloud checks that lists the benchmarks. If the user passes a compliance flag with one of the benchmarks when scanning, only then this check is applied. An example of a check that will not be loaded by default when scanning: https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/aws/cloudwatch/require_cloudtrail_change_alarm.rego#L19-L23. I think we can do something like this here. Wdyt?

simar7:
> @simar7 In such a situation, we specify a frameworks field for cloud checks that lists the benchmarks. If the user passes a compliance flag with one of the benchmarks when scanning, only then this check is applied. An example of a check that will not be loaded by default when scanning: https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/aws/cloudwatch/require_cloudtrail_change_alarm.rego#L19-L23. I think we can do something like this here. Wdyt?

Good point, updated.

checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
```diff
@@ -9,6 +9,9 @@
 # custom:
 #   id: KCV0013
 #   avd_id: AVD-KCV-0013
+#   frameworks:
+#     k8s-cis-1.23:
+#       - "1.23"
```
nikpivkin, line 14: Here you need to specify the ID of the control.
```suggestion
#       - "1.2.13"
```

Also `AVD-KCV-0013` is in `k8s-cis-1.24`.

nikpivkin (changes requested):

checks/kubernetes/cisbenchmarks/apiserver/security_context_deny_plugin.rego
simar7: thanks! updated.

nikpivkin (approved):

PR #281, opened 2024-10-29
Allow the ADD instruction with HTTP, HTTPS and Git URLs

nicwortel:
`COPY` should be preferred over `ADD` when simply copying a file from the build context to the container. However, `ADD` supports additional features such as fetching files from remote HTTP(S) and Git URLS and extracting tar files.

See https://docs.docker.com/build/building/best-practices/#add-or-copy, https://github.com/aquasecurity/trivy/issues/7806 and https://github.com/aquasecurity/trivy/discussions/7791.


nikpivkin (approved):

simar7 (approved):
thanks for the PR!

PR #284, opened 2024-10-30
feat: add check for default network rules in AVD-AZU-0012

nikpivkin:
Related issues:
- https://github.com/aquasecurity/trivy/issues/6459


nikpivkin:
@simar7 I initially thought the default allow rule is created only when public network access is enabled and no user-defined rules exist. However, it is always created whenever no user-defined network rules exist, regardless of public network access settings.

For example, after running the following commands:

- `az storage account create -g trivy-checks-test --name trivycheckstestsa0 --public-network-access 'Disabled'`
- `az storage account update -g trivy-checks-test --name trivycheckstestsa0 --set networkRuleSet='{}'`

the default allow rule still exists.

nikpivkin:
@simar7 Should we check the allow rules in `networkRuleSet` if public access (`publicNetworkAccess`) is disabled?
Since firewall rules only apply to the public endpoint of a storage account, and all network traffic is blocked when public access is disabled, such checks might be redundant.
If public access is intended to be included in the check, I suggest handling it in a separate PR.

simar7:
@nikpivkin is the default value for `default_action` Allow? If so, we would still need to check it in case the user hasn't explicitly set it to Deny.

> However, it is always created whenever no user-defined network rules exist, regardless of public network access settings.


I think this is expected since the `default_action` is what's matched when no other rules are matched.

nikpivkin:
> @nikpivkin is the default value for `default_action` Allow? If so, we would still need to check it in case the user hasn't explicitly set it to Deny.

Yes, it is. But the question was different - does it make sense to check this if public access is disabled? In that case, the firewall rules do nothing.

simar7:
> > @nikpivkin is the default value for `default_action` Allow? If so, we would still need to check it in case the user hasn't explicitly set it to Deny.
> 
> Yes, it is. But the question was different - does it make sense to check this if public access is disabled? In that case, the firewall rules do nothing.

You're right that it wouldn't make sense to check that if public access is disabled but we cannot assume it to be the case.

Are you saying this check should be "skipped" if that's the case? We currently don't have such behavior supported in other checks as we only allow pass or fail for a check unless the user ignores a check. In such a case, we skip it. 

Are there other checks where skipping needs to happen based on certain input scenarios like this one? If so, we might have to think if we should consider adding support for that.

Another alternate approach to handling such a case might be to simply pass the check since it is redundant. This would be inline with Trivy's behaviour to only flag things that require attention by default. WDYT?

ricardo-kh:
Could it be that this change caused the following problem?
https://github.com/aquasecurity/trivy/discussions/9621

PR #295, opened 2024-11-26
feat: add examples of custom checks

nikpivkin:
Related issues: 
- https://github.com/aquasecurity/trivy/issues/4255 

This PR includes:
- examples for almost all kinds of IaC scanners, including the recently added yaml and json.
- using additional data in checks that are passed through the `--config-data` flag
- ignoring built-in checks using Rego
- using built-in Rego functions that Trivy adds. The example uses `sh.parse_commands` to parse shell scripts


nikpivkin:
@simar7 In this PR trivy runs as a cli application via a package, which locks us into the version of trivy specified in go.mod. Alternatively, we can run the latest version of trivy via localstack, which will allow us to run tests in parallel and reduce the number of transitive dependencies (but I don't think this is critical). But then we would have to consider whether it would be too expensive to run such tests often.

simar7:
> But then we would have to consider whether it would be too expensive to run such tests often.

Interesting, why are they expensive to run in localstack? I think running them in localstack is actually good as it adds a layer of integration testing for us in trivy-checks to ensure nothing breaks in Trivy when we release a new trivy-checks version. Also we could add other verisons of Trivy here (besides latest)  to ensure we test against a set of versions.

nikpivkin:
We will pull the Trivy image every time we run the tests . Or can we cache it? If the cache will not be invalidated, we will constantly pull the same version of Trivy in case of using latest tag.

simar7 (approved):

PR #304, opened 2024-12-19
fix: use container as cause in KSV104 check

nikpivkin:
We should only use objects as causes in kubernetes checks , since only they contain the locations used to display the findings.

Config:
```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: hello-sysctls
spec:
  containers:
  - name: hello
    image: busybox
    command:
    - sh
    - "-c"
    - echo 'Hello' && sleep 1h
    securityContext:
      seccompProfile:
        type: Unconfined

```

### Before:
```bash
AVD-KSV-0104 (MEDIUM): container "hello" of pod "hello-sysctls" in "default" namespace should specify a seccomp profile
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
A program inside the container can bypass Seccomp protection policies.

See https://avd.aquasec.com/misconfig/ksv104
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.yaml:1
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 [ ---
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

### After:
```bash
AVD-KSV-0104 (MEDIUM): container "hello" of pod "hello-sysctls" in "default" namespace should specify a seccomp profile
════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
A program inside the container can bypass Seccomp protection policies.

See https://avd.aquasec.com/misconfig/ksv104
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 test.yaml:8-16
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   8 ┌   - name: hello
   9 │     image: busybox
  10 │     command:
  11 │     - sh
  12 │     - "-c"
  13 │     - echo 'Hello' && sleep 1h
  14 │     securityContext:
  15 │       seccompProfile:
  16 └         type: Unconfined
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

### DIFF
--- a/checks/kubernetes/pss/baseline/11_seccomp_profile_unconfined.rego
+++ b/checks/kubernetes/pss/baseline/11_seccomp_profile_unconfined.rego
@@ -44,19 +44,17 @@ container_seccomp_from_annotations(container) := profile {
 
 # containers_with_unconfined_seccomp_profile_type returns all containers which have a seccomp
 # profile set and is profile set to "Unconfined"
-containers_with_unconfined_seccomp_profile_type[name] {
+containers_with_unconfined_seccomp_profile_type[seccomp.container] {
 	seccomp := container_seccomp[_]
 	lower(seccomp.type) == "unconfined"
-	name := seccomp.container.name
 }
 
 # containers_with_unconfined_seccomp_profile_type returns all containers that do not have
 # a seccomp profile type specified, since the default is unconfined
 # https://kubernetes.io/docs/tutorials/security/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads
-containers_with_unconfined_seccomp_profile_type[name] {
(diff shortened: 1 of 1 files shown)


simar7:
In our tests, do we not assert the cause? Should we do so or would that make the tests too brittle?

simar7 (approved):
left a comment but we can merge this in for now.

nikpivkin:
We do not check the cause in Rego tests because the test data is simplified. We can add such a check to the [example tests](https://github.com/aquasecurity/trivy-checks/blob/main/integration/examples_test.go).

PR #307, opened 2024-12-24
fix: align CIDR check rules with their title

nikpivkin:
Related issues:
- https://github.com/aquasecurity/trivy/issues/8184

Checks that require restricting access to or from any IP address now check only this condition. In addition, the titles, descriptions, and messages for all such checks have been unified.

### DIFF
--- a/checks/cloud/aws/ec2/no_public_ingress_acl.rego
+++ b/checks/cloud/aws/ec2/no_public_ingress_acl.rego
@@ -1,5 +1,5 @@
 # METADATA
-# title: Network ACLs should not allow ingress to SSH or RDP from any IP address.
+# title: Network ACLs should not allow unrestricted ingress to SSH or RDP from any IP address.
 # description: |
 #   The Network Access Control List (NACL) function provide stateless filtering of ingress and
 #   egress network traffic to AWS resources. It is recommended that no NACL allows
@@ -56,7 +56,7 @@ deny contains res if {
 	some block in rule.cidrs
 	net.cidr_allows_all_ips(block.value)
 	res := result.new(
-		"Network ACL rule allows ingress from any IP address.",
+		"Network ACL rule allows unrestricted ingress from any IP address.",
 		block,
 	)
 }
--- a/checks/cloud/aws/ec2/no_public_ingress_acl.rego
+++ b/checks/cloud/aws/ec2/no_public_ingress_acl.rego
@@ -1,5 +1,5 @@
 # METADATA
-# title: Network ACLs should not allow ingress from 0.0.0.0/0 to port 22 or port 3389.
+# title: Network ACLs should not allow ingress to SSH or RDP from any IP address.
 # description: |
 #   The Network Access Control List (NACL) function provide stateless filtering of ingress and
 #   egress network traffic to AWS resources. It is recommended that no NACL allows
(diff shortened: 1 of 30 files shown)


itaysk:
@tamirkiviti13 do you have any feedback about this?

tamirkiviti13:
@itaysk 
LGTM, thanks

nikpivkin:
@itaysk @tamirkiviti13 @simar7 The description of some checks (e.g. https://github.com/aquasecurity/trivy-checks/blob/main/checks/cloud/google/compute/no_public_egress.rego) includes the wording `Network security rules should not use very broad subnets`. However, the term `very broad subnets` may not be accurate and may be confusing. When is a subnet considered too broad? This raises questions, as no clear criteria are given in the description.

In addition, there is an inconsistency between the description and the logic of the checks. The description talks about subnets that allow all traffic, i.e. cover the entire range of IP addresses, but in practice the check is triggered if the subnet:
1. Is public.
2. Contains more than one IP address.

Was there a reason why the checks did not originally check subnets containing only the entire IP range? Perhaps the check was intended to be broader to identify other potentially risky configurations, but this should be reflected in the description.

If the purpose of the inspection is to identify subnets that allow all incoming traffic, the logic should be refined to match the description. Otherwise, the description should be adapted to clearly explain the current inspection logic.

This PR contains a fix for only part of the checks.

itaysk (commented):
1. if we change the meaning of some check, should we also change the "short name" etc?
2. do we no longer need the CIDR functions in Go?
3. since we are changing the meaning of some checks, I wonder if we should revisit the severity too.
4. additional discussion about check for unrestricted vs public in the original discussion: https://github.com/aquasecurity/trivy/issues/8184#issuecomment-2593807406

nikpivkin:
1. It seems the new title `should not allow unrestricted ... from the public internet` may be confusing and should be changed to `should not ... from any IP address`. This PR does not change the meaning of the checks but aligns their logic with the title and description.
2. Some checks still use Go functions.
3. Indeed, the severity levels for the checks vary. I found the following levels used: Medium, High and Critical.

simar7:
> if we change the meaning of some check, should we also change the "short name" etc?

short names, or any identifier for that matter should be kept unique and unchanged to ensure we don't cause any unexpected breakage. We would have to deprecate this and introduce a new check if change of names is desired.

checks/cloud/aws/ec2/no_public_ingress_acl.rego
```diff
@@ -1,5 +1,5 @@
 # METADATA
-# title: Network ACLs should not allow ingress from 0.0.0.0/0 to port 22 or port 3389.
+# title: Network ACLs should not allow ingress from the public internet to port 22 or port 3389.
```
simar7, line 2: I think /0 is much more straight forward to understand technically, than the public internet.

nikpivkin: what do you think of “any IP address”?

simar7: Sure that could work.

itaysk: maybe "should not allow unrestricted ingress" ?

nikpivkin: Updated https://github.com/aquasecurity/trivy-checks/pull/307/commits/518ae9c6f53269e2c6a2857a43dfaeb15ef8893e

PR #308, opened 2025-01-04
refactor(repo): Simplify structure

simar7:
- Simplify specs structure
- Document k8s checks

Further this PR classifies k8s checks into:
```
Access: Authentication and authorization related checks (anonymous auth, service accounts)
Network: Network and connection related checks (iptables, streaming connections)
Resources: Admission plugins and resource lifecycle management
Security: Core security features and cryptographic checks
Workloads: Kubelet configuration and operational checks
```


pkg/specs/loader.go
```diff
@@ -10,24 +10,40 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const ComplianceFolder = "compliance"
+// Loader access compliance specs
+type Loader interface {
+	GetSpecByName(name string) string
```
nikpivkin, line 15: Is there a need for this interface?

go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	mvdan.cc/sh/v3 v3.10.0
 )
 
+replace github.com/aquasecurity/trivy => github.com/simar7/trivy v0.0.0-20250104064402-2a2474df782c
```
nikpivkin, line 21: I have opened the [PR](https://github.com/aquasecurity/trivy/pull/8226) . This will help to get rid of the problem due to cyclic dependency.

pkg/specs/loader.go
simar7: Probably not but since we already have it, removal of it would be a breaking change that I don't want to make in this PR.

go.mod
simar7: thanks, updated!

simar7:
@itaysk for your review: do you want to keep node-collector `commands` in this repo? I don't think that feature was fully implemented and as a result there's most likely no consumer of this. I personally would move them out elsewhere as they're not checks per-se (maybe trivy-operator?)

ARCHITECTURE.md
```diff
@@ -7,10 +7,9 @@ This document aims to answer the question *Where is the code that does X?*
 The directory structure is broken down as follows:
 
 - `cmd/` - These CLI tools are primarily used during development for end-to-end testing without needing to pull the library into trivy/tfsec etc.
-- `checks` - All of the checks are defined in this directory.
+- `checks` - All the checks are defined in this directory.
 - `commands` - All Node-collector commands are defined in this directory.
-- `pkg/spec` - Logic to handle standardized specs such as CIS.
-- `pkg/rules` - This package exposes internal rules, and imports them accordingly (see _rules.go_).
+- `pkg/rules` - This package exposes internal checks, and imports them accordingly (see _rules.go_).
```
itaysk, line 12: I don't understand what this means. also 1. didn't we rename rules to checks? 2. shouldn't this code be in trivy?

ARCHITECTURE.md
```diff
@@ -7,10 +7,9 @@ This document aims to answer the question *Where is the code that does X?*
...
 - `specs/` - Standaridized compliance specs such as CIS.
```
itaysk, line 13: I think specs is oerloaded term. for a newcomer "specs" probably relate more to bdd specs or to standardidation specs. the feature in trivy is called "compliance" shouldn't the directory be called that way too?

ARCHITECTURE.md
```diff
@@ -7,10 +7,9 @@ This document aims to answer the question *Where is the code that does X?*
...
 - `test` - Integration tests and other high-level tests that require a full build of the project.
 - `scripts` - Usefule generation scripts for bundle generation and verification purposes.
```
itaysk, line 15: any difference between scripts and cmd? besides the programming languages used

ARCHITECTURE.md
```diff
@@ -7,10 +7,9 @@ This document aims to answer the question *Where is the code that does X?*
 The directory structure is broken down as follows:
 
 - `cmd/` - These CLI tools are primarily used during development for end-to-end testing without needing to pull the library into trivy/tfsec etc.
```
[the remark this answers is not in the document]
itaysk: are any of the tools relevant for an external contributor? if so we should mention them

ARCHITECTURE.md
itaysk: theres a doc about contributing compliance specs under `docs/compliance.md`. it's the only doc in the docs dir I thin we should make it a readme in this dir, like kubernetes related info is documented in readme under checks/kubernetes, then we can remove /docs, which should be in trivy anyway

checks/kubernetes/README.md
```diff
@@ -29,6 +29,19 @@ wget https://github.com/aquasecurity/trivy-checks/raw/main/test/testdata/kuberne
 conftest test denied.yaml --policy myPolicy/ --namespace builtin.kubernetes.KSV008
 ```
 
+# Kubernetes checks classification
+There are several Kubernetes checks that target various subsystems. They are loosely classified as follows.
+
+| Target         | Description                                                                                                                        |
+|----------------|------------------------------------------------------------------------------------------------------------------------------------|
+| Network        | Checks primarily targeting the networking stack                                                                                    |
+| Dynamic        | Checks that evaluate deprecated and removed APIs                                                                                   |
```
itaysk, line 38: either the description is not accurate or the dir name is bad. I think the intention was that these checks rely on environmental context for evaluation. for example, the deprecated API checks rely on information of which k8s version the user is running/evaluating against in order to decide if outdated or not.

checks/kubernetes/README.md
```diff
@@ -29,6 +29,19 @@ wget https://github.com/aquasecurity/trivy-checks/raw/main/test/testdata/kuberne
...
+| CIS Benchmarks | Checks that are recommended by the CIS Benchmarks. The checks inside are targeted per each subsystem (e.g. apiserver, cni, etc.)   |
```
itaysk, line 39: If this dir is serving only CIS, and CIS is using only this dir then fine, but I will challange us abit here: aren't these checks valid k8s checks also outside CIS context? aren't CIS reports include k8s checks besides these checks (i presume something like non privileged pod check).

checks/kubernetes/README.md
```diff
@@ -29,6 +29,19 @@ wget https://github.com/aquasecurity/trivy-checks/raw/main/test/testdata/kuberne
...
+| Advanced       | Checks that are recommended for the advanced uesrs of Kubernetes                                                                   |
```
itaysk, line 40: how do we define advanced users

checks/kubernetes/README.md
```diff
@@ -29,6 +29,19 @@ wget https://github.com/aquasecurity/trivy-checks/raw/main/test/testdata/kuberne
...
+| GKE            | Checks specific to Google Kubernetes Engine                                                                                        |
+| PSS            | Checks pertaining to Pod Security Standards                                                                                        |
```
itaysk, line 42: same comment as CIS

pkg/specs/loader.go
itaysk, line 1: shouldn't this be in trivy?

ARCHITECTURE.md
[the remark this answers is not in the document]
simar7: added

ARCHITECTURE.md
simar7: It's no longer used. I'll remove it.

ARCHITECTURE.md
```diff
@@ -7,10 +7,9 @@ This document aims to answer the question *Where is the code that does X?*
...
-- `checks` - All of the checks are defined in this directory.
+- `checks` - All the checks are defined in this directory.
 - `commands` - All Node-collector commands are defined in this directory.
```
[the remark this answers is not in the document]
simar7: Do we want to continue developing node-collector and its associated functionality? I asked it [here](https://github.com/aquasecurity/trivy-checks/pull/308#issuecomment-2591676880). 

If not we should remove it from trivy-checks if this repo is solely for the purposes of keeping checks.

ARCHITECTURE.md
simar7: fixed.

ARCHITECTURE.md
simar7: No difference besides that. Usually, `cmd/` houses small go programs but that's about it.

checks/kubernetes/README.md
simar7: I agree the name isn't very meaningful but then again I didn't want to change it in this PR as there maybe users who rely on this structure. 

I couldn't come up with a better description, so I'm open to ideas. Would the following work based on what you said?

`"Checks whose decision outcome is dynamically determined based on the environmental context."`

checks/kubernetes/README.md
simar7: You're right in the sense that these checks are ordinary checks even outside of CIS context. 

I'm open to re-organizing all k8s checks in general (maybe based on another type) but we just need to confirm first with the stakeholders that this won't cause any breakage as in the past some users did rely on the structure. Can we do this in another PR?

pkg/specs/loader.go
simar7: Not quite because then users will need to import trivy as a library to load compliance specs which isn't ideal as it's a much bigger dependency.

pkg/spec/spec.go
```diff
@@ -1,23 +0,0 @@
-package spec
```
nikpivkin, line 1: Maybe we should leave this package with public functions, which will be imported from `pkg/compliance` . Then in Trivy you can change the import `spec` to `compliance` and after that completely remove this package? This will get rid of the errors.

ARCHITECTURE.md
itaysk: I see that you've addressed the first comment about renaming specs to compliance but not the second about docs

ARCHITECTURE.md
itaysk: I think scripts is actually mostly about the bundle isn't it? perhaps that would be a better direction

checks/kubernetes/README.md
itaysk: do you think the structure of checks could break users? I mean aren't the checks save/loaded from embedded/bundle in their complete hierarchy? is changing the internal hierachy problemattic?

about the description, it's something like:
"Checks that cannot evaluate based on input alone, and depend on environmental or user-provided context"

checks/kubernetes/README.md
itaysk: yes I think we need to reorganize the checks which is part of the motivation for this PR. sure if you think it needs another pr that's fine

pkg/specs/loader.go
itaysk: users needing to load compliance checks will also need to run them, which will require trivy no? is there a use case for loading compliance and not running them?

nikpivkin: I wonder who else can use the compliance other than Trivy? On the one hand, the compliance is in this repository and if someone else uses it, it is logical to leave the methods to load it here. On the other hand we can do the same thing with compliance as with checks, just import the embedded data.

pkg/spec/spec.go
nikpivkin: @simar7 I've made changes to resolve cyclic dependencies and keep backwards compatibility

ARCHITECTURE.md
nikpivkin: @itaysk The term `compliance spec` and `spec` appears multiple times in the documentation:
- https://github.com/aquasecurity/trivy/blob/main/docs/docs/compliance/compliance.md
- https://github.com/aquasecurity/trivy/blob/main/docs/docs/compliance/contrib-compliance.md

However, its meaning is not explained.

simar7: Should we document `compliance` in this repo? Trivy repo already has docs on it https://github.com/aquasecurity/trivy/tree/7e2a8cb7ad8431ea43da6356769e923d6a52704c/docs/docs/compliance 

>  then we can remove /docs, which should be in trivy anyway

/docs was removed  from trivy-checks (in this PR) so I feel we should update compliance related concerns in Trivy.

ARCHITECTURE.md
simar7: Sure, I simplified it 39d7a8e2424004ea923dcc47d1d547ab44a2924a

ARCHITECTURE.md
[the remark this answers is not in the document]
simar7: Added 5e131f6e3b0e69b090f08d26292a3bc549de7def

checks/kubernetes/README.md
simar7: > I wonder who else can use the compliance other than Trivy?

It's an implementable, exported interface. We have to err on the side of caution as to not inadvertently introduce breakage as it could be used.

checks/kubernetes/README.md
simar7: Sure, as we discussed offline, I've updated: https://github.com/aquasecurity/trivy-checks/pull/308/commits/271a630bfe0f1d7192276c71b2575e2d49b635e0

README.md
```diff
@@ -2,8 +2,18 @@
...
+
+The directory structure is broken down as follows:
+
+- `cmd/` - These CLI tools are primarily used during development for end-to-end testing without requiring the use of a library.
+- `cmd/id` - This command helps generate the next available ID that is free when writing a new check.
+- `checks` - All the checks are defined in this directory.
+- `commands` - All [Node-collector](https://github.com/aquasecurity/k8s-node-collector) commands are defined in this directory.
+- `compliance/` - Standaridized compliance specs such as CIS.
+- `test` - Integration tests and other high-level tests that require a full build of the project.
+- `scripts` - Useful generation scripts for bundle generation and verification purposes.
```
nikpivkin, line 19: This directory no longer exists

nikpivkin, line 19: Should the structure be updated? Right now it does not match the real one.

nikpivkin:
Why were some checks moved to `cloud/kubernetes`? These checks are not related to the cloud provider (e.g. terreaform)

checks/kubernetes/README.md
nikpivkin: I think one check is in the `dynamic` directory because it uses the legacy method of specifying dynamic metadata using the `__rego_input__` rule. Should we use static metadata for this check and specify all dynamic information in the message like we do for all other checks?

README.md
simar7: Thanks for the catch, updated https://github.com/aquasecurity/trivy-checks/pull/308/commits/50398977c0d6c0637dd4586322a7b99bbf8be924

simar7:
> Why were some checks moved to `cloud/kubernetes`? These checks are not related to the cloud provider (e.g. terreaform)

Good catch - It was a mistake, I fixed it here https://github.com/aquasecurity/trivy-checks/pull/308/commits/de6cc940b79e8f330d34b7aa05c84ff46867ead2

nikpivkin:
Put the checks back that were moved in this PR https://github.com/aquasecurity/trivy-checks/pull/344

nikpivkin (approved):

nikpivkin:
Hi @Preisschild !

This seems to be similar to https://github.com/aquasecurity/trivy/issues/8635 . I'll take a look at it.

nikpivkin:
@simar7 I think registry related checks (KSV-32-35) should be disabled by default and enabled on demand, as they conflict with each other. WDYT?

simar7:
> @simar7 I think registry related checks (KSV-32-35) should be disabled by default and enabled on demand, as they conflict with each other. WDYT?

What if we improve the check itself for it to only flag if GCR, ECR or ACR are used? If another registry is used, these checks should not apply.

nikpivkin:
> If another registry is used, these checks should not apply.

Then what will these checks be for? The point of these checks is that they verify that the image is being used from a trusted registry.

itaysk:
@simar7 can you clarify your suggestion? I also didn't get it

@nikpivkin can you clarify how would users enable these if they wanted?

I think this is a case where  the user need to provide some input to the scanner, similar to the `--k8s-version` flag which allows the user to tell the scanner which k8s version to validate against.
Specifically, this check could be generalized into a "trusted registry" check, where the good/bad list comes from user configuration, or rego's `data`, just like the k8s version is implemented. the default configuration can have some examples of trusted registries commented out, and the user can uncomment based on their scenario, or add their own registries. WDYT?

nikpivkin:
> @simar7 can you clarify your suggestion? I also didn't get it
> 
> @nikpivkin can you clarify how would users enable these if they wanted?
> 
> I think this is a case where the user need to provide some input to the scanner, similar to the `--k8s-version` flag which allows the user to tell the scanner which k8s version to validate against. Specifically, this check could be generalized into a "trusted registry" check, where the good/bad list comes from user configuration, or rego's `data`, just like the k8s version is implemented. the default configuration can have some examples of trusted registries commented out, and the user can uncomment based on their scenario, or add their own registries. WDYT?

I think these checks can actually be combined as they all check the registry. The idea of providing trusted registries through data is a good idea. What do you mean by uncommenting? We can have a list of trusted registries and give the user the ability to override them via data.

simar7:
> > @simar7 can you clarify your suggestion? I also didn't get it
> > @nikpivkin can you clarify how would users enable these if they wanted?
> > I think this is a case where the user need to provide some input to the scanner, similar to the `--k8s-version` flag which allows the user to tell the scanner which k8s version to validate against. Specifically, this check could be generalized into a "trusted registry" check, where the good/bad list comes from user configuration, or rego's `data`, just like the k8s version is implemented. the default configuration can have some examples of trusted registries commented out, and the user can uncomment based on their scenario, or add their own registries. WDYT?
> 
> I think these checks can actually be combined as they all check the registry. The idea of providing trusted registries through data is a good idea. What do you mean by uncommenting? We can have a list of trusted registries and give the user the ability to override them via data.

I think we're all talking about the same thing, just differently :)

But yes, what I had in mind is that the user may not agree with what our definition of a "trusted registry" is. In such a case we should  either offer the ability to override this or make this check optional like it was before.

We can start with making it optional first as @nikpivkin pointed out that prior to my PR they were optional. Afterwards, in a different PR we can combine the two checks together into one as suggested.

nikpivkin:
We don't have the option to include checks from the optional directory, does it make sense to move them there?

itaysk:
I don't think combining the checks solves the issue of cloud-specific checks being irrelevant in other cases. I think we need to provide a way to customize the scan to the user's cloud preferences (or more generically to any preferences). I can suggest a feu options:
1. Allow users to select/deselect specific checks in the scan.
    a. possibly already doable with ignore rules? https://github.com/aquasecurity/trivy/discussions/3620
3. Allow users to customize rego data via config. 
    a.   we have discussed per-check configuration https://github.com/aquasecurity/trivy/discussions/4357#discussioncomment-6047605
5. Add a special flag that allows users to specify their clouds and select approproate checks automatically.
    b. we have a similar case for selecting k8s version https://github.com/aquasecurity/trivy/discussions/3620

Discussing a the solution that requires a new feature deserves a proper discussion under Trivy.
In the meantime I suggest that we just disable those checks as a temporary mitigation.

itaysk:
ok, #2 sounds good to me.

PR #310, opened 2025-01-09
refactor(checks): upgrade Rego to v1

nikpivkin:
OPA has been upgraded to v1. The checks are formatted with the `./opa fmt --write --v0-v1 lib checks` command.
```
* 'opa fmt --v0-v1':
  * v0 Rego is formatted to be compatible with v0 AND v1
  * v1 Rego is rejected
```


nikpivkin:
Correct, with this command you can migrate Rego from v0 to v1 and v1 will remain unchanged

nikpivkin:
I added a step in CI to check Rego compatibility with v0 and v1

simar7 (approved):

PR #311, opened 2025-01-10
ci: add Trivy 0.57.1 and 0.58.1 to bundle testing

nikpivkin:


scripts/verify-bundle.go
```diff
@@ -16,7 +16,7 @@ import (
 
 var bundlePath = "bundle.tar.gz"
 var OrasPush = []string{"--artifact-type", "application/vnd.cncf.openpolicyagent.config.v1+json", fmt.Sprintf("%s:application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip", bundlePath)}
-var supportedTrivyVersions = []string{"latest", "canary"} // TODO: add more versions
+var supportedTrivyVersions = []string{"0.57.1", "0.58.1", "latest", "canary"} // TODO: add more versions
```
simar7, line 19: Any particular reason why we should add these two versions in particular?

nikpivkin: I thought we should fix some older versions for backwards compatibility testing, instead of using only the latest ones. I chose some versions starting with 0.57.1 as this is the first version of Trivy where we added bundle v1 support.

simar7: OK let's go ahead with this for now. I wonder if it's worth making this list dynamic in the next iteration of improving this test.

simar7 (approved):

PR #312, opened 2025-01-10
ci: use custom OPA to format Rego

nikpivkin:
I get the following warnings when running `fmt-rego` with OPA v1.0.0 on my machine:
```bash
❯ make fmt-rego
opa fmt -w lib/ checks/ examples/
failed to format Rego source file: 25 errors occurred:
lib/docker/docker.rego:9: rego_parse_error: `if` keyword is required before rule body
lib/docker/docker.rego:9: rego_parse_error: `contains` keyword is required for partial set rules
lib/docker/docker.rego:14: rego_parse_error: `if` keyword is required before rule body
lib/docker/docker.rego:14: rego_parse_error: `contains` keyword is required for partial set rules
lib/docker/docker.rego:19: rego_parse_error: `if` keyword is required before rule body
lib/docker/docker.rego:19: rego_parse_error: `contains` keyword is required for partial set rules
```

So we should use our custom OPA with the fixed version as in other actions

### DIFF
--- a/.github/workflows/test-rego.yaml
+++ b/.github/workflows/test-rego.yaml
@@ -36,18 +36,14 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
-      - name: Setup OPA
-        uses: ./.github/actions/setup-opa
-
       - name: Rego Check
-        run: opa check lib checks --v0-v1
+        run: go run ./cmd/opa check lib checks --v0-v1
 
       - name: OPA Format
         run: |
-          files=$(opa fmt --list . | grep -v vendor || true)
-          if [ -n "$files" ]; then
-            echo "=== The following files are not formatted ==="
-            echo "$files"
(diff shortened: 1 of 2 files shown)


simar7:
Instead of using custom OPA to resolve these issues, should we work towards removing the need for custom OPA altogether?

nikpivkin:
We can't completely do without custom OPA, because we use some custom functions, for example, `result.new`. This also allows us to use the same version of OPA that is used in Trivy.

simar7:
> We can't completely do without custom OPA, because we use some custom functions, for example, `result.new`. This also allows us to use the same version of OPA that is used in Trivy.

Yes I meant that we should maybe find ways to eliminate these custom things so that we can get closer if not at, stock OPA. Doesn't have to be in this PR.

simar7 (approved):

PR #325, opened 2025-01-22
ci: bump Go to 1.23

nikpivkin:


simar7 (approved):
Is there a dependabot hook that we could setup to create a PR for this?

PR #327, opened 2025-01-23
feat: gatekeeper repo ambiguous prefix

itaysk:
Gatekeeper does a prefix-match (startswith) to check if a repo is allowed. this can be leveraged by attackers who create repos with simlar prefix which can bypass the prefix matching (see tests for examples)

related: https://github.com/open-policy-agent/gatekeeper-library/pull/616

### DIFF
--- a/checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
+++ b/checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
@@ -23,7 +23,7 @@ deny contains res if {
 	not contains(repo, ":")
 	res := result.new(
 		"open-ended repository reference in prefix match",
-		repo,
+		input.spec.parameters,
 	)
 }
 
@@ -35,6 +35,6 @@ deny contains res if {
 	count(parts) <= 2
 	res := result.new(
 		"open-ended repository reference in prefix match",
-		repo,
+		input.spec.parameters,
 	)
 }
--- a/checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
+++ b/checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
@@ -5,19 +5,19 @@
 #   - input: schema["kubernetes"]
 # custom:
 #   id: KSV-0124
-#   avdid: AVD-KSV-0124
+#   avd_id: AVD-KSV-0124
 #   severity: HIGH
 package builtin.kubernetes.KSV0124
 
 import rego.v1
 
-relevan_resource if {
+relevant_resource if {
 	input.apiVersion == "constraints.gatekeeper.sh/v1beta1"
 	input.kind == "K8sAllowedRepos"
 }
 
 deny contains res if {
-	relevan_resource
+	relevant_resource
 	some repo in input.spec.parameters.repos
 	not contains(repo, "/")
 	not contains(repo, ":")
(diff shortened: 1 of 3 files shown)


itaysk:
the docs generator didn't pick this up BTW, I'm not sure why

itaysk:
FYI @yakirk

itaysk:
BTW I also wasn't sure if this is a KSV or KUBE check. what's the difference?

checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
```diff
@@ -0,0 +1,40 @@
...
+#   - input: schema["kubernetes"]
+# custom:
+#   id: KSV-0124
+#   avdid: AVD-KSV-0124
+#   severity: HIGH
+package builtin.kubernetes.KSV0124
+
+import rego.v1
+
+relevan_resource if {
```
nikpivkin, line 14:
```suggestion
relevant_resource if {
```

checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
```diff
@@ -0,0 +1,40 @@
+# METADATA
+# title: Gatekeeper repo reference is ambiguously open-ended
+# description: A Gatekeeper policy that references image repositories for prefix-matching is using open-ended and ambiguous pattern, and can potentially match unintended repositories.
+# schemas:
+#   - input: schema["kubernetes"]
+# custom:
+#   id: KSV-0124
+#   avdid: AVD-KSV-0124
```
nikpivkin, line 8:
```suggestion
#   avd_id: AVD-KSV-0124
```

checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
```diff
@@ -0,0 +1,40 @@
...
+}
+
+deny contains res if {
+	relevan_resource
+	some repo in input.spec.parameters.repos
+	not contains(repo, "/")
+	not contains(repo, ":")
+	res := result.new(
+		"open-ended repository reference in prefix match",
+		repo,
```
nikpivkin, line 26: It is better to pass an `input.spec.parameters` object rather than a string, since in k8s only objects contain the location that is required to display the source code.

itaysk: can please you explain this? wdym "in k8s only objects contain the location"

itaysk:
about the docs - this will be covered in a blog post (which is blocked by this PR), i will add a link after it publishes.

checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego
nikpivkin: Trivy exports data to Rego along with metadata that contains the path, file system key and location of the object. Depending on the provider, the exported structure differs. For example for cloud providers all values, even primitives contain metadata, which allows us to accurately display the location of the problem in the report, so in Rego we access it through the `value` field. For k8s the situation is different, when exporting to Rego only objects contain metadata (see https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scanners/kubernetes/parser/manifest_node.go#L56-L62).

itaysk: thanks! I changed the cause accordingly

simar7 (approved):

PR #328, opened 2025-01-27
ci: add lint for checks metadata

nikpivkin:
This PR adds an `invalid_metadata` rule to help maintain consistency, avoid errors, and enforce metadata standards.  This will help to avoid such mistakes: https://github.com/aquasecurity/trivy-checks/pull/327#discussion_r1929999968

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -8,6 +8,8 @@ rules:
         - pattern: '^builtin|lib|defsec|appshield'
           targets:
             - package
+    invalid-metadata:
+      level: error
   naming:
     deny-rule:
       level: error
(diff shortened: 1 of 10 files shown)


simar7 (approved):

simar7:
nice, should we also add a linter for naming of checks? Ref https://github.com/aquasecurity/trivy-checks/pull/327#issuecomment-2622473607

Maybe we allow only certain kinds of checks, based on providers: AVD-AWS-*, AVD-DS-*, etc. So that if there's one that we don't know, we'd have to add it to the lint rule first, if it is a legitimate addition of a new provider.

PR #330, opened 2025-01-30
lint: require input field of Rego metadata

nikpivkin:
Checks that do not use the input field are applied to all input data:
```bash
2025-01-30T02:59:21.0158634Z 2025-01-30T02:55:18Z	WARN	[rego] Module has no input selectors - it will be loaded for all inputs!	file_path="checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego" module="checks/kubernetes/gatekeeper/repo_ambiguous_prefix.rego"
```


simar7 (approved):
thanks!

PR #332, opened 2025-01-30
fix(checks): AVD-AWS-0034 enhanced container insights

MichaelFoleyFZ:
Replaces the current ContainerInsightsEnabled boolean attribute on the ECS cluster settings with ContainerInsightsMode. Container Insights now supports three possible values, enabled, disabled and enhanced (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ClusterSetting.html) 

The enhanced mode is what is recommended by AWS now to get further metrics and details but has been separted from enable/disable due to its cost.

### DIFF
--- a/checks/cloud/aws/ecs/enable_container_insight.rego
+++ b/checks/cloud/aws/ecs/enable_container_insight.rego
@@ -37,9 +37,9 @@ import data.lib.cloud.metadata
 
 deny contains res if {
 	some cluster in input.aws.ecs.clusters
-	not cluster.settings.containerinsightsenabled.value
+	value.is(cluster.settings.containerinsightsmode, "disabled")
 	res := result.new(
 		"Cluster does not have container insights enabled.",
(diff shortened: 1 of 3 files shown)


simar7:
Hmm interesting. So what happens to the old value, was it deprecated?

MichaelFoleyFZ:
> Hmm interesting. So what happens to the old value, was it deprecated?

So the containerInsights setting in the api was always a string but it was just enabled/disabled so fit with being setup as a binary value.

But to add enhanced container insights AWS just added a third possible string value enhanced to the same containerInsights setting

nikpivkin:
Hi @MichaelFoleyFZ !

The rule still checks that `Container Insights` is disabled. So can't we therefore process the string value on the Trivy side and export only the boolean value?

MichaelFoleyFZ:
Will swap the sledge hammer for a screw driver

PR #336, opened 2025-02-15
chore: fix metadata for AVD-KSV-0123

toVersus:
Currently, the metadata for AVD-KSV-0123 is the same as that for [AVD-KSV-01011](https://github.com/aquasecurity/trivy-checks/blob/dbc1975e8fa81ac30489c43cead6dda722615c3a/checks/kubernetes/gke/authenticate_group_bind.rego). Since AVD-KSV-0123 is related to the rule for the `system:masters` group, we will update it to the appropriate info.


nikpivkin (dismissed):
Good catch! LGTM

nikpivkin:
@toVersus Can you run `make docs` and push the changes?

toVersus:
@nikpivkin 
Thanks for your review! Done.

simar7 (approved):

PR #350, opened 2025-03-05
lint: use regal to check duplicate ids

nikpivkin:


test/rules_test.go
```diff
@@ -1,27 +0,0 @@
-package test
-
-import (
-	"testing"
-
-	"github.com/aquasecurity/trivy/pkg/iac/framework"
-	"github.com/aquasecurity/trivy/pkg/iac/rules"
-)
```
simar7, line 8: While it's good that we're removing another dependency on import iac/rules pkg from Trivy but th regal custom checks seem very overly complex to implement to me.

What if we implement this check in Go but instead of importing the iac/rules pkg from Trivy, we instead read the checks from the filesystem?

nikpivkin: I'll take a look.

PR #354, opened 2025-03-06
refactor: use OPA to retrieve checks metadata

nikpivkin:


avd_docs/kubernetes/general/AVD-KSV-0107/docs.md
```diff
@@ -1,13 +0,0 @@
-
-apiVersion '' and kind '' has been deprecated on: '' and planned for removal on:''
```
nikpivkin, line 2: @simar7 The metadata is specified in the __rego__metadata__ rule. I opened a [PR](https://github.com/aquasecurity/trivy-checks/pull/355) with a refactoring of this check.

internal/rego/metadata/metadata.go
```diff
@@ -0,0 +1,80 @@
...
+	case "digitalocean":
+		return "Digital Ocean"
+	case "github":
+		return "GitHub"
+	case "openstack":
+		return "OpenStack"
+	case "cloudstack":
+		return "Cloudstack"
+	default:
+		return cases.Title(language.English).String(strings.ToLower(string(p)))
```
simar7, line 74: Is this cast needed needed? ` cases.Title(language.English)`

simar7 (approved):
lgtm, just left one comment.

internal/rego/metadata/metadata.go
nikpivkin: We lowercase the string first and then capitalize each word. This is taken from Trivy :)

simar7 (approved):

cmd/avd_generator/main.go
```diff
@@ -132,15 +126,15 @@ func fail(msg string, args ...interface{}) {
 }
 
 var docsMarkdownTemplate = `
-{{ .Explanation }}
+{{ .description }}
 
 ### Impact
-{{ if .Impact }}{{ .Impact }}{{ else }}<!-- Add Impact here -->{{ end }}
+{{ if .impact }}{{ .impact }}{{ else }}<!-- Add Impact here -->{{ end }}
```
simar7, line 132: for my knowledge: I thought go templates could access exported fields. I assume these fields are exported, hence this template works as expected but why are they lowercased?

simar7 (approved):
lgtm, just left one comment.

cmd/avd_generator/main.go
nikpivkin: I didn't encounter that the field should be public:
> Templates are executed by applying them to a data structure. Annotations in the template refer to elements of the data structure (typically a field of a struct or a key in a map) to control execution and derive values to be displayed. Execution of the template walks the structure and sets the cursor, represented by a period '.' and called "dot", to the value at the current location in the structure as execution proceeds.

In any case, metadata is a map.

PR #362, opened 2025-03-12
chore: update aws ec2, ecr, ecs, efs examples 

nikpivkin:


simar7:
I'm not sure if I understand but how did you update the examples and what made the diff so big? In some services I see many fields are removed (for instance: https://github.com/aquasecurity/trivy-checks/pull/362/files#diff-68593b7cb3ade47e138e635f50e7515d85e264a3c411cd7c187ff977ae43aa69)

nikpivkin:
The difference is big because of the removal of Go tests, as the test cases are now used as examples. I removed non-misconfiguration fields from the examples to reduce their size and focus the users' attention on the problem itself.

simar7 (approved):

PR #366, opened 2025-03-14
refactor(bundle): Remove bundle manifest

simar7:


simar7:
@nikpivkin this would be a breaking change for older Trivy versions as not having a manifest results in an error as seen in the tests:

```shell
    check_examples_test.go:95: 2025-03-14T02:44:36Z	INFO	[misconfig] Misconfiguration scanning is enabled
        2025-03-14T02:44:36Z	INFO	[misconfig] Need to update the built-in checks
        2025-03-14T02:44:36Z	INFO	[misconfig] Downloading the built-in checks...
        140.81 KiB / 140.81 KiB [------------------------------------------------------] 100.00% ? p/s 100ms2025-03-14T02:44:36Z	ERROR	[misconfig] Falling back to embedded checks	err="check load error: manifest file open error (/root/.cache/trivy/policy/content/.manifest): open /root/.cache/trivy/policy/content/.manifest: no such file or directory"
```

Fortunately our bundle is versioned so we could update our bundle major version, but doing that to me as the only change in the version is a bit excessive. Maybe we simply wait before we remove the manifest from the bundle, or do you have any other ideas?

nikpivkin:
@simar7 I don't think this change gives us anything and so it can wait. We can come back to it when we are going to release the next major version of the bundle.

nikpivkin:
The way the bundle is built has changed, so I opened a new PR https://github.com/aquasecurity/trivy-checks/pull/485

PR #369, opened 2025-03-14
fix: check only managed load balancers

nikpivkin:
Trivy assigns orphaned load balancer listeners (those that do not have an associated load balancer) to a dummy load balancer. Such a dummy load balancer should not be checked.


simar7 (dismissed):

simar7:
Can we add tests in the respective checks for this case?

nikpivkin:
> Can we add tests in the respective checks for this case?

Done https://github.com/aquasecurity/trivy-checks/pull/369/commits/50b8b2c592f311b3538a999cb5d7a3daf92a68e6

PR #373, opened 2025-03-19
feat(checks): Restrict s3 from wild card access

simar7:
## Description

S3 buckets are often used by other AWS services as a data store. Keeping only essential access and limiting any wide open access to the buckets is critical for avoiding S3 access abuse.

Closes https://github.com/aquasecurity/trivy/issues/8671


checks/cloud/aws/iam/no_policy_wildcards.rego
```diff
@@ -29,3 +29,5 @@
 #             provider: aws
 #   examples: checks/cloud/aws/iam/no_policy_wildcards.yaml
 package builtin.aws.iam.aws0057
+
+# TODO(simar7): This checks seems to have lost its flesh during the transition.
```
simar7, line 33: @nikpivkin I found this check got lost in the transition. We should add it back.

simar7: Ah I just realized this is a deprecated check 😄

simar7:
@nikpivkin how can we check for specific ARNs if they are referenced in the check rather than defined. See example below of an input which should trigger this check but doesn't because the ARN is of an AWS policy not a custom one.

```terraform
# Provider configuration
provider "aws" {
  region = "us-west-2"
}

# Create an IAM role
resource "aws_iam_role" "example_role" {
  name = "example-role"

  assume_role_policy = jsonencode({
    Version = "2012-10-17",
    Statement = [
      {
        Effect = "Allow",
        Principal = {
          Service = "ec2.amazonaws.com"
        },
        Action = "sts:AssumeRole"
      }
    ]
  })
}

# Attach the AmazonS3FullAccess policy to the IAM role
resource "aws_iam_role_policy_attachment" "example_role_policy_attachment" {
  role       = aws_iam_role.example_role.name
  policy_arn = "arn:aws:iam::aws:policy/AmazonS3FullAccess"
}
```

Here `"arn:aws:iam::aws:policy/AmazonS3FullAccess"` is essentially the following:

```json
{
  "Version" : "2012-10-17",
  "Statement" : [
    {
      "Effect" : "Allow",
      "Action" : [
        "s3:*",
        "s3-object-lambda:*"
      ],
      "Resource" : "*"
    }
  ]
}
```

Ref: https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonS3FullAccess.html

nikpivkin:
To do this, we need to download AWS managed policies and either generate a Rego file or import them as data. This increases maintenance overhead, so we have dropped support for `no_policy_wildcards.rego` check.

[Here](https://github.com/zoph-io/MAMIP/tree/master/policies) is the repository containing these policies.

nikpivkin:
Would you be able to add a description of why this check is necessary and why only for S3 access checks?

simar7:
What if we take a simpler approach and statically define the ARNs which we are looking for? For instance, `"arn:aws:iam::aws:policy/AmazonS3FullAccess"` is one of them. Then we could loop through this list of ARNs in this check to see if we find any.

simar7:
> Would you be able to add a description of why this check is necessary and why only for S3 access checks?

This check is looking for potentially wide open S3 buckets that are used by other services as a data source.

nikpivkin:
> What if we take a simpler approach and statically define the ARNs which we are looking for

We won't be able to keep the ARN list up to date, but this approach is simple enough so we can start there.

checks/cloud/aws/iam/limit_s3_full_access_test.rego
```diff
@@ -0,0 +1,53 @@
+package builtin.aws.iam.aws0345
+
+import rego.v1
+
+test_with_allow_s3_full_access if {
+	policies := [{
+		"name": "policy_with_s3_full_access",
+		"document": {"value": "{\"Version\":\"2012-10-17\",\"Id\":\"\",\"Statement\":[{\"Sid\":\"\",\"Effect\":\"Allow\",\"Principal\":{},\"NotPrincipal\":{},\"Action\":[\"s3:*\"],\"NotAction\":null,\"Resource\":[\"*\"],\"NotResource\":null,\"Condition\":{}}]}"},
```
nikpivkin, line 8: You can use the [json.marshal](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-encoding-jsonmarshal) function to make policies more readable.

simar7: thanks, updated.

nikpivkin (approved):

PR #380, opened 2025-04-05
fix(checks): Update description of KSV037

simar7:
Fixes: https://github.com/aquasecurity/trivy/issues/8691


nikpivkin:
The short code still contains a mention of pods: `short_code: no-user-pods-in-system-namespace`. It should be added to the aliases and this field should be updated.

nikpivkin (commented):
The short code still contains a mention of pods: `short_code: no-user-pods-in-system-namespace`. It should be added to the aliases and this field should be updated.

nikpivkin (approved):

PR #386, opened 2025-04-09
chore: pass schemas to opa test

nikpivkin:


nikpivkin:
This PR will be ready for review when we remove Trivy from the dependencies.

simar7:
interesting, what does passing the schema to the tests do differently? Is it able to type check?

nikpivkin:
Yes, passing the schema to the tests performs type checking. This works the same way as it does now inside Trivy by passing the schema in compile checks. Right now the schema is checked by Trivy itself in some Go tests. But in the future, when we remove Trivy from dependencies, we will need to pass the schema to OPA for type checking.

nikpivkin:
It may be better to pass the schema to `opa check`.

simar7 (approved):

PR #391, opened 2025-04-18
feat: combine multiple image registry checks into one

nikpivkin:
Related issue: https://github.com/aquasecurity/trivy/issues/8728


nikpivkin:
@simar7 How about standardising the path to the custom data in all checks and documenting this? Checks can import data using the path `data.{id}.{name}`. This would get rid of name conflicts in the future and improve the readability of the data files.

I find this easier to maintain:
```yaml
id001:
  foo: foo
id002:
  bar: bar
```

Than this:
```yaml
foo: foo
bar: bar
```

checks/kubernetes/security/uses_untrusted_registry.rego
```diff
@@ -0,0 +1,96 @@
...
+gcr_registries := {
+	"gcr.io",
+	"us.gcr.io",
+	"eu.gcr.io",
+	"asia.gcr.io",
+}
+
+default_trusted_registries := (azure_registries | ecr_registries) | gcr_registries
+
+all_trusted_registires := ksv0125.trusted_registries if {
```
simar7, line 76:
```suggestion
all_trusted_registries := ksv0125.trusted_registries if {
```

simar7:
> @simar7 How about standardising the path to the custom data in all checks and documenting this? Checks can import data using the path `data.{id}.{name}`. This would get rid of name conflicts in the future and improve the readability of the data files.
> 
> I find this easier to maintain:
> 
> ```yaml
> id001:
>   foo: foo
> id002:
>   bar: bar
> ```
> 
> Than this:
> 
> ```yaml
> foo: foo
> bar: bar
> ```
> 
> Custom registries can be specified in `ksv0125.trusted_registries` for this check.

Yeah that's a good call. Should we also add this to a regal check, if writing a regal check is not too overly complicated for this? 😄

nikpivkin:
> Yeah that's a good call. Should we also add this to a regal check, if writing a regal check is not too overly complicated for this? 😄

Ok, let me make this a separate PR as there are a few more checks worth documenting. I'll take a look, the rule shouldn't be complicated.

simar7:
> > Yeah that's a good call. Should we also add this to a regal check, if writing a regal check is not too overly complicated for this? 😄
> 
> Ok, let me make this a separate PR as there are a few more checks worth documenting. I'll take a look, the rule shouldn't be complicated.

OK sounds good - we would just need to rebase this PR.

checks/kubernetes/uses_untrusted_registry_test.rego
```diff
@@ -0,0 +1,66 @@
+package builtin.kubernetes.KSV0125_test
+
+import data.builtin.kubernetes.KSV0125 as check
+
+import rego.v1
+
+test_check_registry[name] if {
```
nikpivkin, line 7: By the way, these are parameterised tests for which OPA has recently added support. See https://www.openpolicyagent.org/docs/policy-testing#parameterized-tests-and-data-driven-testing

nikpivkin:
@simar7 Created the [PR](https://github.com/aquasecurity/trivy-checks/pull/415) with a custom rule to check data import.

simar7 (approved):

nikpivkin:
> How about standardising the path to the custom data in all checks and documenting this?

@simar7 Do you have any ideas where we should put the documentation?

nikpivkin:
@itaysk Should we document information about the custom data in this check and a few others in this repository or Trivy? I will create a PR tomorrow.

PR #399, opened 2025-04-26
feat: add check to detect URL annotations with suspicious characters

nikpivkin:
Related issues:
- https://github.com/aquasecurity/trivy/issues/8672


checks/kubernetes/network/suspicious_url_annotation_test.rego
```diff
@@ -0,0 +1,41 @@
...
+		"unsupported annotation": {
+			"key": "foo",
+			"value": "http://example.com/#;\ninjection_point",
+			"expected": 0,
+		},
+		"alb oidc": {
+			"key": "alb.ingress.kubernetes.io/auth-idp-oidc",
+			"value": `'{"issuer":"https://example.com","authorizationEndpoint":"http://example.com/#;\ninjection_point","tokenEndpoint":"https://token.example.com","userInfoEndpoint":"https://userinfo.example.com","secretName":"my-k8s-secret"}'`,
+			"expected": 1,
+		},
```
simar7, line 28: note: since we don't have a  way to definitively prove an annotation key would have a value of a URL, we are relying here for two specific keys:

1. `"alb.ingress.kubernetes.io/auth-idp-oidc"` which is part of the kubernetes well known URL annotations
2. `"link.argocd.argoproj.io/external-link"` a commonly used annotation in argoCD.

determining if a key will  have a value of a URL is a hard problem. therefore we've started with well-known annotations for now. 

WDYT @itaysk?

itaysk: I guess that is fine. we could also search for regexp matches across all annotations but I imagine the concern was for performance?

simar7: Performance certainly yes but there could be annotations where a URL is expected as a value but the key may not make it obvious as such.

PR #406, opened 2025-05-07
Fix unreachable URL of AVD-AWS-0320 metadata

nekketsuuu:
<https://docs.aws.amazon.com/AmazonS3/latest./dev/transfer-acceleration.html> is not found, whereas <https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html> is redirected to <https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration.html>.


simar7 (approved):

nekketsuuu:
Thanks!

PR #410, opened 2025-05-12
fix: correctly detect empty objects in KSV118

nikpivkin:
Detecting an empty object in the source configuration by comparing it to `{}` is incorrect, since an empty object with no entries will contain metadata.

For example, a `securityContext:` object would be exported to Rego as:
```json
{
  "securityContext": {
    "__defsec_metadata": {...}
  }
}
```


simar7:
Should we  add a test for this case?

nikpivkin:
> Should we add a test for this case?

Added https://github.com/aquasecurity/trivy-checks/pull/410/commits/dfa2f8321904102fe43d3deb89b32e05331c5242

simar7 (approved):

PR #411, opened 2025-05-12
add example check for terraform

nikpivkin:
Related PRs:
- https://github.com/aquasecurity/trivy/pull/8741

### DIFF
--- a/examples/terraform-raw/main.tf
+++ b/examples/terraform-raw/main.tf
@@ -1,0 +1,7 @@
+resource "aws_s3_bucket" "test" {
+  bucket = "test"
+  tags = {
+    Environment = "Production"
+    Project     = "ProjectX"
+  }
+}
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,13 +1,11 @@
 module github.com/aquasecurity/trivy-checks
 
-go 1.24
-
-toolchain go1.24.0
+go 1.24.2
 
 require (
-	github.com/aquasecurity/trivy v0.61.1-0.20250407075540-f1329c7ea1aa
+	github.com/aquasecurity/trivy v0.62.1-0.20250528213022-4a7ebb70b48f
```
simar7, line 6: @nikpivkin I've updated to the latest trivy but yet in the CI it's complaining about lack of the `raw-config-scanners` flag. Any ideas why?

It briefly worked for me locally prior to rebasing the branch. Not sure why would that cause this to happen.

simar7: I see - yeah I think we can merge and cut a release then.

simar7 (approved):

PR #414, opened 2025-05-17
refactor(checks): Flatten all kubernetes checks

simar7:


nikpivkin:
@simar7 Why was it decided to simplify the structure of the `kubernetes` directory?

simar7:
> @simar7 Why was it decided to simplify the structure of the `kubernetes` directory?

The directories prior were subjectively decided by me based on subsystems they touched. After reviewing it with @itaysk we concluded that it might be better to drop  such classification as it's not obvious to the external user to determine which subsystem the check would fall under. There were also many checks that could arguably overlap with each other, hence the change.

nikpivkin (approved):

PR #415, opened 2025-05-21
chore: add regal rule to check data import prefix

nikpivkin:


nikpivkin:
Such a structure is created by `regal` https://docs.styra.com/regal/custom-rules#creating-a-new-rule . I've simplified it https://github.com/aquasecurity/trivy-checks/pull/415/commits/f21daf01dcb8c38e7ffc43f14d7a0a4e2880c393

PR #420, opened 2025-05-22
lint: enable 'duplicate-rule' rule

nikpivkin:
https://docs.styra.com/regal/rules/bugs/duplicate-rule

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -16,3 +16,6 @@ rules:
       ignore:
         files:
           - "*_test.rego"
+  bugs:
+    duplicate-rule:
+      level: error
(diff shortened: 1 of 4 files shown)


simar7 (approved):
good catch - on this topic, does rego allow polymorphism? I read the doc and it states it only considers a duplicate if the name is the same.

nikpivkin:
> rego allow polymorphism

Functions declared in rego take values of dynamic type, so you can write the following:
```rego
foo(arg) if {
  is_array(arg)
  ...
}

foo(arg) if {
  is_string(arg)
  ...
}
```

> it only considers a duplicate if the name is the same.

If the names are different, they are no longer duplicates, are they? Or did I misunderstand.

PR #421, opened 2025-05-24
docs: Add a note for cost consideration for AVD-AWS-0090 aws-s3-enable-versioning

nekketsuuu:
As AWS recommends, it's better to utilize S3 Lifecycle rules to optimize cost for S3 buckets with versioning enabled. Since this is primarily for cost optimization rather than security, we decided to just leave a note about this in the document.

closes https://github.com/aquasecurity/trivy/discussions/8885


simar7 (approved):

nekketsuuu:
Thanks!

PR #430, opened 2025-05-31
BREAKING: refactor(checks): Remove deprecated ID field from checks

simar7:


nikpivkin:
Should id values be added to alias to avoid breaking ignore rules for users?

simar7:
> Should id values be added to alias to avoid breaking ignore rules for users?

AFAIK the legacy ID (e.g. XYZ-1234, where XYZ is a service) field was never used in the checks. As for long id, we could add that as an alias.

nikpivkin:
Yes indeed, ID is not used in the ignore rules: https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scan/result.go#L269-L283

simar7:
> Yes indeed, ID is not used in the ignore rules: https://github.com/aquasecurity/trivy/blob/main/pkg/iac/scan/result.go#L269-L283

Added the long  IDs as aliases eaa64d6e61347331684e3e861da33b55f6712016

nikpivkin (approved):

nikpivkin:
@simar7 Did you use a script or agent to add the aliases?

simar7:
> @simar7 Did you use a script or agent to add the aliases? That's a big chore :)

Yes perfect task for an LLM! 

Actually there are some aliases that don't quite follow the convention of provider-service-check but I had to keep them in as removing them will incur breakage.

PR #436, opened 2025-06-08
feat: Add comprehensive Talos Linux support to node collector configu…

tijmenvandenbrink:
…ration

- Add Talos-specific CA file paths (/system/secrets/kubernetes/)
- Add Talos service configuration paths (/system/services/)
- Add Talos kubeconfig paths for kubelet and kube-proxy
- Add Container Runtime Interface configuration paths
- Extends existing partial Talos support with complete path coverage

This improves node-collector's ability to audit Talos Linux clusters by including Talos-specific file system paths alongside existing systemd-based paths for other distributions.

Note: The /etc/systemd mkdir issue on Talos still requires removing systemd volume mounts from trivy-operator configuration since Talos uses a different init system.

### DIFF
--- a/commands/config/node.yaml
+++ b/commands/config/node.yaml
@@ -73,6 +73,8 @@ node:
       - /var/lib/rancher/rke2/agent/server.crt
       - /var/lib/rancher/rke2/agent/client-ca.crt
       - /var/lib/rancher/k3s/agent/client-ca.crt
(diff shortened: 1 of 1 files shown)


martinholovsky:
Please review, many of us have rootfs as read-only and Talos doesn't use systemd at all - it has its own custom init system.

PR #441, opened 2025-06-23
refactor(misconf): Use `id` and `long_id` for misconfig checks

simar7:


simar7:
@nikpivkin I updated the IDs to conform to XXX-0123, where XXX is the service. 

I believe the integration tests are expected failures (correct me if I'm wrong).

Would you know why the regal linter is failing?

nikpivkin:
We decided to keep `long_id` as a human-readable format, but it builds on `short_code`, which is removed.

nikpivkin:
> > We decided to keep `long_id` as a human-readable format, but it builds on `short_code`, which is removed.
> 
> Right - I updated that [757f4e7](https://github.com/aquasecurity/trivy-checks/commit/757f4e7083c084c5b4b124db65d406758456ebeb)

Long ID has the following format: `<provider>-<service>-<short-code>` :)

nikpivkin:
What if we move the past value of `AVD-ID` (and maybe `ID`) in alias to the first value? Then it won't break the rules of ignoring users and tests.

simar7:
> > > We decided to keep `long_id` as a human-readable format, but it builds on `short_code`, which is removed.
> > 
> > 
> > Right - I updated that [757f4e7](https://github.com/aquasecurity/trivy-checks/commit/757f4e7083c084c5b4b124db65d406758456ebeb)
> 
> Long ID has the following format: `<provider>-<service>-<short-code>` :)

Thanks I missed that on other clouds besides aws. It should be updated now https://github.com/aquasecurity/trivy-checks/pull/441/commits/3cbd7124c3854ca3d01be9beef7993ca316ac26a  - but if you see something else got missed, let me know. As for docker and kubernetes, we don't have any subcategories anymore so it's simply `docker/kubernetes-checkname` for the long_id for now.

simar7:
> What if we move the past value of `AVD-ID` (and maybe `ID`) in alias to the first value? Then it won't break user ignore rules and tests.

Yeah that would work the best but the challenge remains how to get the aliases updated across the checks with older IDs (AVDID and ID).

One idea I had in mind is to load up checks via the library and then run a script that uses those IDs to update the rego files in place. I didn't try this yet but if you want to give it a shot, go ahead.

nikpivkin (dismissed):

simar7:
> @simar7 I have made the following changes to this PR:
> 
> * Updated Kubernetes check IDs to follow the `CODE-NUMBER` format (e.g., `KSV-0024`).
> * Added all values from `id`, `avd_id`, and `short_code` fields to aliases, avoiding duplicates when `id` and `avd_id` were the same.
> * Updated `long_id` values to use dashes (`-`) instead of underscores (`_`).
> * Updated the Regal rule to include validation for the `long_id` field.
> * Updated integration tests to use aliases for assertions, since the `avd_id` field (previously used as the report ID in JSON output) has been removed.
> 
> I didn't pull the last changes and accidentally erased the changes made in the last 3 commits related to the checks, but have already put everything back in.

@nikpivkin I have tried to rebase it over latest main (as you know we rolled back the ID changes previously merged). Hopefully I didn't overlook something while rebasing, so could you take a look to see if everything is in place?

nikpivkin:
> So could you take a look to see if everything is in place?

Yeah, sure.

nikpivkin:
@simar7 [Diff](https://github.com/aquasecurity/trivy-checks/compare/48485c238616d7ee383f319612187f17fde18a2c..607f7a2454f663a1d41d0c375b896bcedf523acf) contains no changes, so it's okay

PR #457, opened 2025-07-17
Revert "refactor(misconf): Use `id` and `long_id` for misconfig checks (#441)"

simar7:
This reverts commit 2b5288202a3f94c7ce831f5eded6ae37ac77b354.


nikpivkin:
> > > @nikpivkin for some reason `AVD-GCP-0071` and `AVD-GCP-0072` are failing integration tests, would you know why?
> > 
> > 
> > I discovered that these checks use incorrect port fields — `fromport` and `toport` instead of the correct `start` and `end`, which caused the integration tests to fail. Recently, I added a test that validates Rego test input data against the schema, but it only verified field types and didn't catch unexpected fields. I opened a [PR](https://github.com/aquasecurity/trivy-checks/pull/458) to improve this test by patching the schema to add `additionalProperties: false`, which disallows any undocumented fields.
> 
> @nikpivkin I just rebased on the latest main with your changes but it seems `AVD-GCP-0072` still seems to be red.

I noticed that after merging the PR with the ID change, tests only started running for a small subset of checks. This PR fixes that :) . So the tests did not catch the incorrect behaviour of the recently added AVD-GCP-0072 check. I opened [PR](https://github.com/aquasecurity/trivy-checks/pull/461) to fix the check. I also opened a PR in Trivy, but there is no need to wait for it as the PR in trivy-checks is sufficient to fix it.

nikpivkin:
@simar7 I opened a new PR because it's difficult to track changes in this one. https://github.com/aquasecurity/trivy-checks/pull/490

PR #458, opened 2025-07-18
test: disallow extra fields in cloud schema validation

nikpivkin:
This PR adds the `additionalProperties` property (set to `false`) to all Cloud objects in the schema, except for `iac.types.MapValue`, which is a map and is expected to accept arbitrary keys. This makes the validation stricter and helps catch unexpected fields in test inputs.

### DIFF
--- a/checks/cloud/google/sql/encrypt_in_transit_data_test.rego
+++ b/checks/cloud/google/sql/encrypt_in_transit_data_test.rego
@@ -17,11 +17,11 @@ test_deny_tls_not_required if {
 	count(res) == 1
 }
 
-test_deny_tls_not_required_but_ssl_mode_require_ssl if {
-	inp := build_input({"requiretls": {"value": false}, "ssl_mode": {"value": check.ssl_mode_trusted_client_certificate_required}})
+test_allow_tls_not_required_but_ssl_mode_require_ssl if {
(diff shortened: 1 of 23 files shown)


checks/cloud/google/sql/encrypt_in_transit_data_test.rego
```diff
@@ -17,11 +17,11 @@ test_deny_tls_not_required if {
 	count(res) == 1
 }
 
-test_deny_tls_not_required_but_ssl_mode_require_ssl if {
-	inp := build_input({"requiretls": {"value": false}, "ssl_mode": {"value": check.ssl_mode_trusted_client_certificate_required}})
+test_allow_tls_not_required_but_ssl_mode_require_ssl if {
+	inp := build_input({"requiretls": {"value": false}, "sslmode": {"value": check.ssl_mode_trusted_client_certificate_required}})
 
 	res := check.deny with input as inp
-	count(res) == 1
```
nikpivkin, line 24: This test was incorrect and should have revealed that the check allows such input data.

nikpivkin:
All test fixes are related to input data not matching the schema, except for one. I've left a comment above explaining that case.

simar7 (approved):

PR #459, opened 2025-07-21
feat: Add 2 storage checks

yagreut:
Added 2 gcp storage checks:
Enable_bucket_logging- GCP-0077
Enable_bucket_versioning- GCP-0078

### DIFF
--- a/checks/cloud/google/storage/enable_bucket_logging.rego
+++ b/checks/cloud/google/storage/enable_bucket_logging.rego
@@ -32,8 +32,14 @@ import rego.v1
 
 import data.lib.cloud.metadata
 
+buckets_for_logging := {name |
+	some bucket in input.google.storage.buckets
+	name := bucket.logging.logbucket.value
+}
+
 deny contains res if {
 	some bucket in input.google.storage.buckets
+	not bucket.name.value in buckets_for_logging
 	not bucket.logging.logbucket.value
 
 	res := result.new(
--- a/avd_docs/google/storage/GCP-0077/Terraform.md
+++ b/avd_docs/google/storage/GCP-0077/Terraform.md
@@ -3,6 +3,7 @@ Enable Access and Storage logs for Cloud Storage buckets by configuring a log si
 
 
 ```hcl
+# Bucket with logging configured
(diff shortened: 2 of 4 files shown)


checks/cloud/google/storage/enable_bucket_logging.rego
```diff
@@ -0,0 +1,42 @@
...
+#   - https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#log_bucket
+# custom:
+#   id: GCP-0077
+#   aliases:
+#     - google-storage-cloud-storage-bucket-logging-not-enabled
+#     - AVD-GCP-0077
+#   long_id: google-storage-enable-bucket-logging
+#   provider: google
+#   service: storage
+#   severity: MEDIUM
```
nikpivkin, line 18: Since this check uses an updated schema, the `minimum_trivy_version` field should be added so that this check is skipped in older versions of Trivy. See https://trivy.dev/latest/docs/scanner/misconfiguration/custom/#customavd_id-and-customid. I'll update the integration tests to skip checks for older versions of trivy that don't yet support this field.

nikpivkin: Same for the other check.

simar7:
> Since a scheme from Trivy is now used to validate the input test data, we should first wait for the merge of [this PR](https://github.com/aquasecurity/trivy/pull/9226).

Right - it's nice to see this in action!

checks/cloud/google/storage/enable_bucket_logging.rego
simar7: @nikpivkin one test that comes to my mind are the integration tests where we run the checks bundle against specific trivy versions. In such a case, we could use the minimum trivy version info from a check to only run it against the versions it targets. 

That way, if a check author specifies an incorrect minimum trivy version, it will the integration test as the support will not be present in the older (unsupported) version of trivy.

I feel check authors (albeit a few) may not realize what the minimum trivy version to target would be.

nikpivkin (approved):
LGTM!

PR #462, opened 2025-07-24
test: skip checks by minimum_trivy_version

nikpivkin:


nikpivkin:
> Should adding a `minimum_trivy_version` be strictly enforced? So far it's been an optional field targeting new checks where the schema updates might be necessary but I wonder if it's small enough of a hurdle to enforce every time we write a new check.

Determining the minimum required Trivy version for a check is not a trivial task. It depends on the latest version in which changes were made to the schema that affect references to object fields within the check. This can be partially detected by running integration tests against different Trivy versions, as we are currently doing.

If a check references fields that are not present in older versions, compilation will fail at the type-checking stage. However, due to current limitations in OPA’s type system (as I mentioned earlier), such issues may not always be accurately detected. If OPA’s type-checking improves in the future, we will be able to rely on it more confidently for version compatibility.

> Besides the overhead of knowing what minimum trivy version will support a check, are there other downsides of such a mandate?

I don't see

simar7 (approved):

PR #464, opened 2025-08-01
fix(gcp): properly check if bucket has logging

nikpivkin:
The expression `not bucket.logging.logbucket.value` is true when the `value` field is missing, but by default Trivy fills the `logbucket` with empty values:

```json
"logging": {
...
              "logbucket": {
                "endline": 6,
                "explicit": false,
                "filepath": "main.tf",
                "fskey": "fd569cbde7899a2c12c7fa8a4c51c3777becfc1797289d3e1c9e960fd56802a2",
                "managed": true,
                "resource": "google_storage_bucket.default",
                "sourceprefix": "",
                "startline": 1,
                "unresolvable": false,
                "value": ""
              },
...
```

which is why this expression is never true:
```
| | Fail not bucket.logging.logbucket.value
```

Fixes:
https://github.com/aquasecurity/trivy-checks/actions/runs/16662841182/job/47163530682

### DIFF
--- a/checks/cloud/google/storage/enable_bucket_logging.rego
+++ b/checks/cloud/google/storage/enable_bucket_logging.rego
@@ -40,10 +40,12 @@ buckets_for_logging := {name |
 deny contains res if {
 	some bucket in input.google.storage.buckets
 	not bucket.name.value in buckets_for_logging
-	not bucket.logging.logbucket.value
+	not has_logging(bucket)
 
 	res := result.new(
 		"Storage bucket logging is not configured with a target log bucket.",
 		metadata.obj_by_path(bucket, ["logging", "logbucket"]),
 	)
 }
+
+has_logging(bucket) if bucket.logging.logbucket.value != ""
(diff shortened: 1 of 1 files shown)


nikpivkin:
@simar7  This error was not caught on the canary version in CI because Trivy applied a version filter and skipped this check. We should think about how to test the checks in such a situation.

nikpivkin:
@simar7 I raised the minimum required version of Trivy to 0.66.0 due to incorrect field handling in Trivy. See https://github.com/aquasecurity/trivy/pull/9296

simar7:
> @simar7 This error was not caught on the canary version in CI because Trivy applied a version filter and skipped this check. We should think about how to test the checks in such a situation.

>  but by default Trivy fills the logbucket with empty values:

I agree - I think the absence of an attribute is difficult to articulate since it could be due to the user forgetting (actual misconfiguration) or a bug (like in this case). 

I made a suggestion here https://github.com/aquasecurity/trivy/pull/9296/files#r2249131532

simar7 (approved):
LGTM - I left a comment.

PR #479, opened 2025-09-16
fix(docker): DS-0001 handle ARG-based FROM; warn on unresolved ARG; deny implicit latest; keep explicit :latest

black-leg-nameko:
## Summary

This PR enhances the `DS-0001` rule (`docker-use-specific-tags`) for Dockerfiles to properly handle **ARG-based `FROM`** instructions.

**Problem:** `FROM ${BASE}` could bypass `:latest` detection when the ARG is unresolved or when the tag is implicitly omitted after substitution.

**What this PR does:**
- Emits a **warning** when `FROM` contains **unresolved build ARGs** at scan time.
- Denies **implicit `latest`** (tag omitted) after ARG resolution.
- Keeps denying **explicit `:latest`** (existing behavior).
- Still allows **digest-pinned** images (`@sha256:...`).

This closes the coverage gap without increasing false positives in CI.

---

## Changes

- `checks/docker/latest_tag.rego`
  - Added helpers:
    - `all_var_refs_resolvable(s, vars)` to detect unresolved `${VAR}` / `$VAR`
    - `digest_pinned_from(instruction)` to treat `@sha256:` as safe
  - New rules:
    - **warn**: unresolved ARGs in `FROM`
    - **deny**: implicit `latest` (no tag) after resolving variables
    - **deny**: explicit `:latest` (existing rule retained; message clarified)
- `checks/docker/latest_tag_test.rego`
  - Added tests for:
    - ARG with implicit latest (deny)
    - ARG with explicit `:latest` (deny)
    - ARG unresolved (warn)
    - ARG with digest pin (allowed)
- `checks/docker/latest_tag.yaml`
  - Updated recommended action to clarify ARG-based cases and digest pinning.

---

## Testing Evidence

Local repro set under `tmp/test-trivy`:

- `Dockerfile.arg-implicit`  
  - **MEDIUM**: “Omitted tag in the 'FROM' statement for image 'ubuntu'; pin version or digest.”  
  - **MEDIUM**: “Specify a tag in the 'FROM' statement for image 'ubuntu'”  
  - (plus existing checks: USER, HEALTHCHECK)

- `Dockerfile.arg-latest`  
  - **MEDIUM**: “Omitted tag in the 'FROM' statement for image 'nginx'; pin version or digest.”  
  - **MEDIUM**: “Specify a tag in the 'FROM' statement for image 'nginx'”  
  - (plus existing checks: USER, HEALTHCHECK)

- `Dockerfile.unresolved-arg`  
  - **WARN** expected: “FROM uses unresolved build ARGs: … Ensure the resolved image is version- or digest-pinned.”  
  - Note: depending on output format, WARN may not appear in the “Failures” summary, but is emitted by the rule.

- `Dockerfile.digest`  
  - DS-0001 family: **no findings** (digest-pinned is allowed)

Also verified with `--config-check` in dev namespace while developing to avoid package collisions; final rule keeps the original `builtin.dockerfile.DS001`.

---

## Rationale (Why warn, not deny, for unresolved ARGs)

Unresolved build ARGs depend on the build-time environment. Hard-failing them would create noisy false positives in CI. A **warning** surfaces the risk while keeping pipelines usable, and the deny paths still catch both explicit and implicit `latest`.

---

## Backward Compatibility

- Existing explicit `:latest` detection remains intact.
- Digest-pinned images are unaffected.
- No provider/service IDs were changed; only Dockerfile rule logic extended.

### DIFF
--- a/checks/docker/latest_tag.rego
+++ b/checks/docker/latest_tag.rego
@@ -99,6 +99,58 @@ is_string_ending_with_var(reference) if {
 	endswith(reference, var_ref)
 }
 
+
+# FROM with digest pinning is considered safe
+digest_pinned_from(instruction) if {
+    contains(instruction.Value[0], "@sha256:")
+}
+
+# Returns true if all ${VAR}/$VAR in s can be resolved from vars
+all_var_refs_resolvable(s, vars) if {
+    refs := find_var_refs(s)
+    count(refs) == 0
+} else if {
+    every ref in find_var_refs(s) {
+        object.get(vars, extract_var_name(ref), null) != null
+    }
+}
+
+# unresolved build args should not be hard failed, but surfaced as a warning
+warn contains res if {
+    some instruction in input.Stages[i].Commands
+    instruction.Cmd == "from"
+    contains(instruction.Value[0], "$")
+    vars := stage_vars(i)
+    not all_var_refs_resolvable(instruction.Value[0], vars)
+    res := result.new(
+        sprintf("FROM uses unresolved build ARGs: %q. Ensure the resolved image is version- or digest-pinned.", [instruction.Value[0]]),
+        instruction,
+    )
+}
+
+# Implicit latest (tag omitted) after resolving variables -> deny
+deny contains res if {
+    some instruction in input.Stages[i].Commands
+    instruction.Cmd == "from"
+    not digest_pinned_from(instruction)
+
+    vars := stage_vars(i)
+    [img, tag] := parse_image_and_tag(instruction, vars)
+
+    img != "scratch"
+    img != ""
+    not is_alias(img)
+
+    tag == "latest"
+    not endswith(lower(instruction.Value[0]), ":latest")
+
+    res := result.new(
+        sprintf("Omitted tag in the 'FROM' statement for image '%s'; pin version or digest.", [img]),
+        instruction,
+    )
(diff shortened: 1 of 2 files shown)


checks/docker/latest_tag.rego
```diff
@@ -99,6 +99,58 @@ is_string_ending_with_var(reference) if {
...
+    refs := find_var_refs(s)
+    count(refs) == 0
+} else if {
+    every ref in find_var_refs(s) {
+        object.get(vars, extract_var_name(ref), null) != null
+    }
+}
+
+# unresolved build args should not be hard failed, but surfaced as a warning
+warn contains res if {
```
nikpivkin, line 119: Trivy doesn’t have a separate concept of “warnings” in checks — instead, it uses severity with different levels. We intentionally chose not to notify users about unresolved Dockerfile ARGs, since Trivy performs static analysis while those arguments are only resolved during image build.

black-leg-nameko:
Thanks for the feedback and clarification!  
I apologize for not following the usual workflow of opening a Discussion first.  
I now understand the reasoning behind not reporting unresolved ARGs, and I appreciate the explanation.  

I’ll close this PR for now and will start with a Discussion next time to align with maintainers before implementing changes.  
Thanks again for your guidance!

PR #487, opened 2025-09-29
feat: add example for custom check on raw Terraform config

nikpivkin:


examples/terraform-raw/required-aws-tags.rego
```diff
@@ -0,0 +1,49 @@
...
+import rego.v1
+
+resource_types_to_check := {"aws_s3_bucket"}
+
+resources_to_check := {block |
+	some module in input.modules
+	some block in module.blocks
+	block.kind == "resource"
+	block.type in resource_types_to_check
+}
```
simar7, line 26: WDYT if we add some helper rego functions to export such fields easily? That way the user could avoid a lot of repetitive code.

nikpivkin: Yes, I thought about that when I was working on the scheme for nested blocks  in https://github.com/aquasecurity/trivy/pull/9570.

PR #488, opened 2025-09-30
test: use ORAS image from GHCR

nikpivkin:
Bitnami [has discontinued support](https://github.com/bitnami/containers#%EF%B8%8F-important-notice-upcoming-changes-to-the-bitnami-catalog) for several images. As a result, the `bitnami/oras` image on Docker Hub no longer has any available tags. It has now been replaced with `ghcr.io/oras-project/oras` from GHCR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ docs-test:
 
 .PHONY: create-bundle
 create-bundle:
-	go run ./cmd/bundle -root .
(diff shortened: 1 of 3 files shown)


simar7 (approved):

nikpivkin:
> There might be other projects in our org that could be using the same image.

or other images from bitnami

PR #527, opened 2026-02-07
fix(k8s): correct RKE2 CIS 1.24 audit paths

T1mn:
Fixes #10167

The audit definitions contained incorrect paths for RKE2 platform. All commands specifying RKE2 platform required path updates per the official RKE2 CIS 1.24 documentation.

Path corrections:
- /node/var/lib/etcd → /var/lib/rancher/rke2/server/db/etcd (etcd)
- /node/etc/kubernetes/ssl → /var/lib/rancher/rke2/server/tls (PKI)
- /etc/kubernetes/admin.conf → /var/lib/rancher/rke2/server/cred/admin.kubeconfig

Platform separation:
- Removed rke2 from 6 k8s commands (admin.conf, controller-manager.conf, scheduler.conf permissions/ownership)
- Created 6 RKE2-specific commands (CMD-0051 through CMD-0056) with proper paths and unique keys
- Fixed kubePKIKeyFilePermissionsRancher.yaml platform: rke → rke2
- Updated rke2-cis-1.24.yaml compliance spec with new command IDs

### DIFF
--- a/commands/kubernetes/adminConfFilePermissionsRancher.yaml
+++ b/commands/kubernetes/adminConfFilePermissionsRancher.yaml
@@ -1,0 +1,8 @@
+---
+- id: CMD-0051
+  key: adminConfFilePermissions-rke2
+  title: admin.conf file permissions
+  nodeType: master
+  audit: stat -c permissions=%a /var/lib/rancher/rke2/server/cred/admin.kubeconfig
+  platforms:
+    - rke2
--- a/commands/kubernetes/controllerManagerConfFileOwnershipRancher.yaml
+++ b/commands/kubernetes/controllerManagerConfFileOwnershipRancher.yaml
@@ -1,0 +1,8 @@
+---
+- id: CMD-0054
+  key: controllerManagerConfFileOwnership-rke2
+  title: controller-manager.conf file ownership
+  nodeType: master
+  audit: stat -c %U:%G /var/lib/rancher/rke2/server/cred/controller.kubeconfig
+  platforms:
+    - rke2
--- a/commands/kubernetes/schedulerConfFileOwnershipRancher.yaml
+++ b/commands/kubernetes/schedulerConfFileOwnershipRancher.yaml
@@ -1,0 +1,8 @@
+---
+- id: CMD-0056
+  key: schedulerConfFileOwnership-rke2
+  title: scheduler.conf file ownership
+  nodeType: master
+  audit: stat -c %U:%G /var/lib/rancher/rke2/server/cred/scheduler.kubeconfig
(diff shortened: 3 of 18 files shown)


commands/kubernetes/adminConfFilePermissionsRancher.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- id: CMD-0051
+  key: adminConfFilePermissions-rke2
+  title: admin.conf file permissions
+  nodeType: master
+  audit: stat -c permissions=%a /var/lib/rancher/rke2/server/cred/admin.kubeconfig
```
afdesk: I'm unsure, that we should use `permissions=%a` here.
`kube-bench` checks a flag "permissions", but `trivy-checks` uses a policy `builtin.kubernetes.KCV0060`:
https://github.com/aquasecurity/trivy-checks/blob/a2cb1cce27aeed7c528a33e6545eef2e1ba1c599/checks/kubernetes/apiserver_admin_conf_permission.rego#L31

commands/kubernetes/controllerManagerConfFileOwnershipRancher.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- id: CMD-0054
+  key: controllerManagerConfFileOwnership-rke2
+  title: controller-manager.conf file ownership
+  nodeType: master
+  audit: stat -c %U:%G /var/lib/rancher/rke2/server/cred/controller.kubeconfig
```
afdesk: about hardcoded paths. 
there was the same issue in `kube-bench`: https://github.com/aquasecurity/kube-bench/issues/1843
it was resolved here: https://github.com/aquasecurity/kube-bench/pull/1844/

Shouldn't we go the same way here?
i mean to update https://github.com/aquasecurity/trivy-checks/blob/main/commands/config/node.yaml

afdesk (changes requested):

commands/kubernetes/schedulerConfFileOwnershipRancher.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- id: CMD-0056
+  key: schedulerConfFileOwnership-rke2
+  title: scheduler.conf file ownership
+  nodeType: master
+  audit: stat -c %U:%G /var/lib/rancher/rke2/server/cred/scheduler.kubeconfig
```
nikpivkin: The issue of paths has already been raised above. Here, it is most likely possible to update `node.yaml` and use `scheduler.kubeconfig`.

PR #602, opened 2026-08-18
fix(AWS-0112): allow TLS 1.3 and enhanced security policies

sadvi11:
AWS-0112 compared the SAM API domain security policy against `TLS_1_2` exactly, so every enhanced policy that negotiates TLS 1.3 was reported as outdated. It now accepts a set of policies, the same way `aws-elasticsearch-use-secure-tls-policy` already does.

The set is an allow list rather than a deny on `TLS_1_0` so that a policy AWS introduces later is reported until it has been reviewed, instead of being silently accepted. A test covers that case.

The values are taken from the `SecurityPolicy` allowed values documented for `AWS::ApiGateway::DomainName`, which is where SAM passes this property through to. Of the eleven documented values, only `TLS_1_0` permits TLS 1.0.

### Relates issues:
- https://github.com/aquasecurity/trivy/issues/11117

### DIFF
--- a/checks/cloud/aws/sam/api_use_secure_tls_policy.rego
+++ b/checks/cloud/aws/sam/api_use_secure_tls_policy.rego
@@ -30,7 +30,6 @@ package builtin.aws.sam.aws0112
 import rego.v1
 
 import data.lib.cloud.metadata
-import data.lib.cloud.value
 
 deny contains res if {
 	some api in input.aws.sam.apis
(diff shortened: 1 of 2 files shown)


nikpivkin:
Closing this as a duplicate of #600, which was opened first for aquasecurity/trivy#11117. Before opening a PR, please check the existing open PRs and leave a comment on the issue that you are working on it, as described in our contributing guidelines.

PR #605, opened 2026-08-22
fix: use glob patterns for TLS policies in AWS-0126

Dev9269:
Fixes #11118

## Changes
- Replace hardcoded policy list with glob patterns to match all TLS 1.2+ policies (Policy-Min-TLS-1-2-*, Policy-Min-TLS-1-3-*)
- Add alue.is_known guard to skip unresolvable values
- Fixes false positive for Policy-Min-TLS-1-2-RFC9151-FIPS-2024-08 (TLS 1.3 with FIPS)
- Follows the same pattern used in AWS-0005 for API Gateway security policies

## Testing
- Added test case for FIPS policy (	est_allow_use_secure_tls_policy_fips)
- Added test case for unresolvable values (	est_allow_unresolvable_tls_policy)
- Updated YAML examples with FIPS policy as good example

### DIFF
--- a/checks/cloud/aws/apigateway/use_secure_tls_policy.rego
+++ b/checks/cloud/aws/apigateway/use_secure_tls_policy.rego
@@ -30,13 +30,12 @@ package builtin.aws.apigateway.aws0005
 
 import rego.v1
 
+import data.lib.aws.apigateway as apigateway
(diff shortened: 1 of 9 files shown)


Dev9269:
Superseded by #607 (AWS-0126) and #606 (AWS-0112)

PR #608, opened 2026-08-23
fix(elasticsearch): use pattern matching for TLS policy check

lxcxjxhx:
## Description

Fix false positive in AWS-0126 when the TLS policy is `Policy-Min-TLS-1-2-RFC9151-FIPS-2024-08` (the strictest FIPS-compliant policy).

## Problem

The check only accepted two hard-coded TLS policy names:
- `Policy-Min-TLS-1-2-2019-07`
- `Policy-Min-TLS-1-2-PFS-2023-10`

This caused `Policy-Min-TLS-1-2-RFC9151-FIPS-2024-08` (TLS 1.3 with FIPS) to be reported as insecure, which is a false positive.

Additionally, the check lacked a `value.is_known` guard, so unresolvable values also produced findings.

## Fix

1. **Use glob patterns** instead of exact names: `["Policy-Min-TLS-1-2-*", "Policy-Min-TLS-1-3-*"]`
   - Covers all current secure policies
   - Automatically covers future TLS 1.2+ and 1.3+ policies
   - Correctly rejects `Policy-Min-TLS-1-0-2019-07`

2. **Add `value.is_known` guard** to prevent false positives on unresolvable values

## Testing

Added test cases for:
- `Policy-Min-TLS-1-2-RFC9151-FIPS-2024-08` (should pass)
- `Policy-Min-TLS-1-3-2025-01` (should pass, future TLS 1.3 policy)
- `Policy-Min-TLS-1-0-2019-07` (should fail, insecure)
- Unresolvable values (should fail)

## Related Issues

Fixes #11118

### DIFF
--- a/checks/cloud/aws/elasticsearch/use_secure_tls_policy.rego
+++ b/checks/cloud/aws/elasticsearch/use_secure_tls_policy.rego
@@ -40,9 +40,12 @@ deny contains res if {
 	)
 }
 
-recommended_tls_policies := {
-	"Policy-Min-TLS-1-2-2019-07",
-	"Policy-Min-TLS-1-2-PFS-2023-10",
-}
+# Patterns rather than exact names, so policies added to the same families
+# are covered without a change here.
+secure_tls_policies := ["Policy-Min-TLS-1-2-*", "Policy-Min-TLS-1-3-*"]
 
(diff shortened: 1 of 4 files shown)


nikpivkin:
Closing this as a duplicate of #598, which was opened first for aquasecurity/trivy#11118. Before opening a PR, please check the existing open PRs and leave a comment on the issue that you are working on it, as described in our contributing guidelines.

PR #609, opened 2026-08-23
fix(sam): use pattern matching for TLS policy check

lxcxjxhx:
## Description

Fix false positive in AWS-0112 when using newer TLS policies like SecurityPolicy_TLS13 and SecurityPolicy_TLS12 EDGE.

## Problem

The check only accepted the literal value TLS_1_2 as a secure policy, rejecting all other secure policies.

## Fix

Use the same pattern matching approach as AWS-0005 (API Gateway):
- TLS_1_2 (exact match)
- SecurityPolicy_TLS12_*_EDGE (pattern)
- SecurityPolicy_TLS13_* (pattern)

This covers all 12 valid policies and automatically covers future policies of the same families.

Also add value.is_known guard to prevent false positives on unresolvable values.

## Related Issues

Fixes #11117

### DIFF
--- a/checks/cloud/aws/elasticsearch/use_secure_tls_policy.rego
+++ b/checks/cloud/aws/elasticsearch/use_secure_tls_policy.rego
@@ -40,9 +40,12 @@ deny contains res if {
 	)
 }
 
-recommended_tls_policies := {
-	"Policy-Min-TLS-1-2-2019-07",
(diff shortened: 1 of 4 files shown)


nikpivkin:
Closing this as a duplicate of #600, which was opened first for aquasecurity/trivy#11117. Before opening a PR, please check the existing open PRs and leave a comment on the issue that you are working on it, as described in our contributing guidelines.

PR #611, opened 2026-08-31
fix(AWS-0126): allow secure TLS policies and ignore unresolvable values

pranavsivaprasad230501:
The check only accepted two exact TLS policy names, so the newer and stricter `Policy-Min-TLS-1-2-RFC9151-FIPS-2024-08` policy was reported as outdated. It also lacked a `value.is_known` guard, so an unresolved value produced a finding.

Match on `Policy-Min-TLS-1-2-*` and `Policy-Min-TLS-1-3-*` patterns instead of an exact list, so future policies in the same families are covered without a further change, and guard on `value.is_known`, matching the pattern already used by `AWS-0005` and the recently-fixed Azure/GCP TLS checks (`AZU-0026`, `GCP-0039`).

Closes aquasecurity/trivy#11118.

## Testing

- Added test cases for the previously-false-flagged FIPS policy and for an unresolvable value, alongside the existing test cases.
- Added a `good` Terraform example demonstrating the FIPS policy and regenerated the docs.
- `make rego` passes clean (fmt, schema check, lint, tests, docs) — 1945/1945 tests pass.

### DIFF
--- a/avd_docs/aws/elasticsearch/AWS-0126/Terraform.md
+++ b/avd_docs/aws/elasticsearch/AWS-0126/Terraform.md
@@ -9,6 +9,14 @@ resource "aws_elasticsearch_domain" "good_example" {
   }
 }
 ```
+```hcl
+resource "aws_elasticsearch_domain" "good_example_fips" {
+  domain_endpoint_options {
+    enforce_https       = true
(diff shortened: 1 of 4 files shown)


nikpivkin:
Closing this as a duplicate of #598, which was opened first for aquasecurity/trivy#11118. Before opening a PR, please check the existing open PRs and leave a comment on the issue that you are working on it, as described in our contributing guidelines.

PR #617, opened 2026-09-04
refactor(rego): move the result.new and isManaged built-ins from Trivy

nikpivkin:
Trivy imports this repository for the embedded checks, the compliance specs and the Rego built-ins. This repository imported Trivy back, only so that `result.new` and `isManaged` are registered and the checks can be compiled and tested here. Because of that the module graph went in a circle. A trivy-checks bump in Trivy pulled a requirement on an older Trivy back into the graph, and a Trivy bump here pulled in an older trivy-checks the same way. The whole Trivy dependency tree also came here just for these two functions.

Neither built-in uses anything from Trivy. Both only build an object out of the resource metadata, so they belong next to the checks that call them. Now this repository registers them itself and does not depend on Trivy at all.


internal/rego/result.go
```diff
@@ -0,0 +1,124 @@
...
+		resultKeyManaged:      ast.BooleanTerm(true),
+		resultKeyFSKey:        ast.StringTerm(""),
+		resultKeyResource:     ast.StringTerm(""),
+		resultKeyParent:       ast.NullTerm(),
+	}
+	if msg != nil {
+		metadata[resultKeyMessage] = msg
+	}
+
+	// A dockerfile command carries the location in its own fields, not in metadataKey.
```
DmitriyLewen: `internal/rego/result.go:76` calls the fallback branch dockerfile-specific, but most cloud checks go through it as well: `BaseValue[T].ToRego()` inlines the metadata next to `value`, with no `__defsec_metadata`.
The ratio of `result.new` calls in this repository: 451 in `checks/cloud`, 174 in `checks/kubernetes`, 32 in `checks/docker`.
The `dockerKey*` constants and the comments on lines 29 and 98 are accurate — line 76 is the only one that is off.
A small thing, but the code is moving to another repository and this comment becomes the only context.

```go
// The metadata is either nested under metadataKey or inlined in the object itself:
// structs use metadataKey, while wrapped values and dockerfile commands keep the
// location in their own top-level fields.
input := cmp.Or(cause.Get(ast.StringTerm(metadataKey)), cause)

File: internal/rego/result_test.go
Comment (DmitriyLewen) [inline internal/rego/result_test.go:1]: The four cases in this file cover an empty input, `__defsec_metadata`, a dockerfile command and `parent` — but not the flat shape, which is the dominant path.
That does not look like a coincidence: the missing test and the inaccurate comment cover for each other.

Suggested case to add:

```go
		{
			name: "metadata inlined in a wrapped value",
			cause: `{
	"value": false,
	"startline": 12,
	"endline": 14,
	"sourceprefix": "",
	"filepath": "main.tf",
	"explicit": false,
	"managed": true,
	"unresolvable": false,
	"fskey": "fs-key",
	"resource": "aws_s3_bucket.example"
}`,
			expected: `{
	"msg": "message",
	"startline": 12,
	"endline": 14,
	"sourceprefix": "",
	"filepath": "main.tf",
	"explicit": false,
	"managed": true,
	"fskey": "fs-key",
	"resource": "aws_s3_bucket.example",
	"parent": null
}`,
		},
```

I checked it: it passes against the current code and `gofmt` is clean.
And it is not a tautology — break the fallback (replace the second argument of `cmp.Or` with an empty object) and the case fails.

It also pins down the fact that `unresolvable` is not carried from the input into the result: the field is produced by `Metadata.ToRego()`, but neither `result.new` nor `parseCause` reads it, and checks take it straight from the input instead (`lib/cloud/value.rego`).
That is long-standing behaviour and does not need changing in this PR, but having it fixed by a test keeps the next reader from mistaking it for an oversight.

nikpivkin: added https://github.com/aquasecurity/trivy-checks/pull/617/changes/f615cb4aad6d797476a61130a061db15dd6c4757

internal/rego/result.go
nikpivkin: At first, I copied the comment exactly as it was. Now I've updated it so that it reflects the current behavior. https://github.com/aquasecurity/trivy-checks/pull/617/changes/f615cb4aad6d797476a61130a061db15dd6c4757

knqyf263:
The direction looks good to me. It might be worth adding depguard to prevent trivy-checks from accidentally importing Trivy again. Since golangci-lint isn't currently set up here, that could take some extra work, so I'll leave it up to you whether to include it here or handle it separately.

DmitriyLewen (dismissed):

nikpivkin:
> The direction looks good to me. It might be worth adding depguard to prevent trivy-checks from accidentally importing Trivy again. Since golangci-lint isn't currently set up here, that could take some extra work, so I'll leave it up to you whether to include it here or handle it separately.

Since part of the code in this repo is written in Go, I think it would be helpful to add golangci-lint. I'll do that in another PR.

knqyf263 (dismissed):

afdesk (approved):

PR #624, opened 2026-09-08
refactor(rego): iterate the shell AST with syntax.Preorder

nikpivkin:
Collecting commands from a parsed shell script needed a `syntax.Walk` callback that returned `true` on every node, so the traversal was spread across a closure and the real work was a single type switch inside it.

`mvdan.cc/sh/v3` v3.14 adds `syntax.Preorder`, an iterator over all nodes, so the same traversal is now a plain loop over the AST. The version bump also brings parser fixes for shell that we could not parse before, such as `#` inside `[[ ]]`.


afdesk (dismissed):

nikpivkin:
@afdesk I rebased the branch and did `go mod tidy`

